PR #9, opened 2016-12-22
Switch to geometry

stirringhalo:


stirringhalo:
Note to anyone reading, this may be a breaking change if you don't have all your docker images pulling latest at this point in time, specifically: v1.2 on import-natural-earth and latest on openmaptiles-tools.

lukasmartinelli:
That's just the default. We should still probably set the `geometry_field` to geometry.

PR #15, opened 2017-04-09
Add ZRes support to Openmaptiles-tools

stirringhalo:
This pull adds support for Z## in the tolerance field, which is then turned into a real value and put into the final tile schema. 
Has some error checking for bad values.
Needs a final test of functionality.


stirringhalo:
Needs testing to see if tolerance isn't already an int/float.

PR #23, opened 2017-10-11
Fix for Generated Metadata for bounds and centered always compute to zero (`0`)

roblabs:
`generate-metadata` is required when generating mbtiles for use with `tileserver-gl`.  It generates a valid `version`, `attribution` and others, but it does not properly compute the center.  `tileserver-gl` relies on the center to give a proper starting point and to generate valid thumbnails 

This approach uses `docopt` to pass in a bounding box.  

TODO — parse os.environ.get('BBOX') for west, south, east, north, values

---

#### Current defaults, average computes to zero 
https://github.com/openmaptiles/openmaptiles-tools/blob/37c9e052a9de40476f68252e6bd13185e2c2462e/bin/generate-metadata#L100-L103

----

### Example with this PR

generate-metadata

### DIFF
--- a/bin/generate-metadata
+++ b/bin/generate-metadata
@@ -3,6 +3,7 @@
 
 Usage:
   generate_metadata <mbtiles>
+  generate_metadata <mbtiles> <west> <south> <east> <north>
   generate_metadata (-h | --help)
   generate_metadata --version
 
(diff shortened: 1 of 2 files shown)


roblabs:
The documentation and scripts would also need to be updated

For example, the line in quickstart.sh 
https://github.com/openmaptiles/openmaptiles/blob/2b2051c0858eee08a80c6aab36f4a11397cf2904/quickstart.sh#L268

``` 
docker-compose run --rm openmaptiles-tools  generate-metadata ./data/tiles.mbtiles
```

Would need to change to 
```
docker-compose run --rm openmaptiles-tools  generate-metadata ./data/tiles.mbtiles -180 -85 180 85
```

MartinMikita:
Thanks @roblabs for your PR!
We are merging it, and will modify to parse BBOX, if args are not passed.

PR #31, opened 2019-02-13
Install docker-compose as part of the docker image

nyurik:
Implements https://github.com/openmaptiles/openmaptiles/issues/583

This patch allows any make command via a docker, without any extra packages on the host:
```
docker run \
  -v $(pwd):/tileset \
  -v /var/run/docker.sock:/var/run/docker.sock \
  openmaptiles/openmaptiles-tools \
  make refresh-docker-images
```

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -2,6 +2,7 @@ FROM python:3.6
 
 RUN apt-get update \ 
     && apt-get install  -y --no-install-recommends \
+        docker-compose \
         graphviz \
         sqlite3 \
     && rm -rf /var/lib/apt/lists/


MartinMikita:
No. Run make refresh docker images does not need to be run via docker openmaptoles-tools.
It will not work, as you run docker in docker

nyurik:
@MartinMikita this is not correct - it is possible to run docker in docker (which is not the same as running docker `daemon` inside itself). The `-v /var/run/docker.sock:/var/run/docker.sock` parameter makes it possible - the docker inside docker communicates with the host's docker daemon, thus making it possible, while also not requiring anything other than the docker itself to be installed on the host (e.g. without the `make` utility, etc). Please take a look at [this post](https://jpetazzo.github.io/2015/09/03/do-not-use-docker-in-docker-for-ci/)

MartinMikita:
As declared in this comment - https://github.com/openmaptiles/openmaptiles/issues/583#issuecomment-475535900, we are not willing to support docker-in-docker as it is not required by the project itself.

Thanks for your comments and understanding.

PR #34, opened 2019-03-22
SQL code to generate MVT with a single gettile(z,x,y) call

nyurik:
This change adds a `generate-sqltomvt` script that can be used to generate:
* an SQL prepared statement  `gettile(z,x,y)` (3 integers) -- query returns a single row with a single column -- the Map Vector Tile (MVT) as a binary blob
* a create statement for a full-fledged PostgreSQL function to do the same

The generated SQL code uses hardcoded extent of 4096 and tile width/height of 256.
It would be relatively easy to parametrize it if required.

Optionally, `generate-sqltomvt` supports empty tile detection -- if the tile has just a single layer (e.g. "water"), and the zoom level is higher than `--mask-level` (default=8), and a square (0,0 to 4096,4096) is fully within the tile's water layer geometry, no rows will be returned.

This code can be used directly by the new version of the postserve server -- see corresponding PR https://github.com/openmaptiles/postserve/pull/11


nyurik:
To discuss this and other OpenMapTiles issues, please join [OSM Slack](https://osmus-slack.herokuapp.com/) - **#openmaptiles** channel.

PR #47, opened 2019-10-18
Water detection - use ST_COLLECT instead of ST_UNION

nyurik:
This should avoid many errors of this type:

```
10/593/222:  GEOSUnaryUnion: TopologyException: Input geom 0 is invalid: Self-intersection at or near point 1930 1039 at 1930 1039
```


nyurik:
this will need to be reworked further -- ST_COLLECT is not correct, and ST_UNION fails on some paths, investigating...

PR #59, opened 2019-10-24
Import tilebbox from vt-util

eyusupov:
See https://github.com/openmaptiles/import-sql/pull/13


nyurik:
I'm really confused why would we need this?  `TileBBox()` is already included as part of [postgis-vt-util](https://github.com/mapbox/postgis-vt-util/blob/v1.0.0/postgis-vt-util.sql#L488). We may want to upgrade to the latest version (v1.1.0), but that's a separate discussion.

eyusupov:
You're right, I was using old docker image, thank you

PR #61, opened 2019-10-25
Use Postgres 12

eyusupov:


nyurik:
Thanks @eyusupov , I just realized we should use `$(source /etc/os-release && echo $VERSION_CODENAME)` instead of `buster` in the `deb http://apt.postgresql.org/pub/repos/apt/ buster-pgdg main $pg_major` line. I will fix it in a separate PR.

nyurik:
oh, and it should have been `$PG_MAJOR`. Strangly enough it still works.

Issue #57, opened 2019-10-24
Make Postserve watch for SQL file changes


nyurik:
When a layer SQL file changes, make Postserve automatically start using it.  If prepared statement is used (preferable), Postserve should close and re-open the connection.

PR #62, opened 2019-10-25
Use connection pool and asyncpg

eyusupov:
Due to connection pool it's no longer convenient to use prepared
statements, which are per-session, so a stored procedure is used
instead.

See https://github.com/openmaptiles/postserve/pull/15


nyurik:
I really like the faster code, but I don't think we should switch to stored procedures because that means all of the server's users must use the same UDF, and the server must allow write access -- neither of which is true in many cases.

Each user should be able to experiment with their own SQL changes without modifying the server (moreover, in a proper environment the server might even be read-only, or you could connect to a replica or even a pool of replicas).

Unrelated but nice to have -- see https://github.com/openmaptiles/openmaptiles-tools/issues/57 -- probably a separate PR though.

openmaptiles/postserve.py
```diff
@@ -148,12 +123,22 @@ def serve(port, pghost, pgport, dbname, user, password, metadata, tileset_path,
...
     ])
 
     application.listen(port)
     print(f"Postserve started, listening on 0.0.0.0:{port}")
     print(f"Use http://localhost:{port} as the data source")
-
     tornado.ioloop.IOLoop.instance().start()
+
+
+def m():
```
nyurik: this file is not the entry point - we use [bin/postserve](https://github.com/openmaptiles/openmaptiles-tools/blob/master/bin/postserve) for that, and I don't think we should have two entry points for the same program.

eyusupov:
I've updated the code to use the query instead of functions. According to https://github.com/MagicStack/asyncpg/issues/51, they are still going to be cached by asyncpg underneath. Also, I've noticed that I was running my tests with the wrong container, and the original code was quite broken.Sorry for that.

bin/postserve
```diff
@@ -14,6 +14,7 @@ Options:
   --mask-layer=<layer>  If tile only has this layer covering for the whole tile (e.g. 'water'),
                         and requested zoom is more than --mask-zoom parameter, returns empty result
   --mask-zoom=<z>       If --mask-layer is set, tiles will be checked for empty after given zoom [default: 8]
+  -s --serve=<port>     Return this hostname as tileserver URL in [default: localhost]
```
nyurik: missing space on this and next line caused the `default` feature not to work. Adding it makes the code below redundant.
```suggestion
  -s --serve=<port>     Return this hostname as tileserver URL in  [default: localhost]
```

bin/postserve
```diff
@@ -14,6 +14,7 @@ Options:
...
   -p --port=<port>      Serve on this port [default: 8090]
```
nyurik:
```suggestion
  -p --port=<port>      Serve on this port  [default: 8090]
```

bin/postserve
```diff
@@ -65,7 +66,8 @@ if __name__ == '__main__':
         metadata['maskLevel'] = mask_zoom
 
     serve(
-        port=int(args['--port']),
+        host=args.get('--serve', 'localhost'),
```
nyurik: `get` not needed if you add an extra space above.
```suggestion
        host=args['--serve'],
```

bin/postserve
```diff
@@ -65,7 +66,8 @@ if __name__ == '__main__':
...
+        port=int(args.get('--port', '8090')),
```
nyurik:
```suggestion
        port=int(args['--port']),
```

PR #64, opened 2019-10-26
support for generating SQL field mappings based on layer definitions

nyurik:
see README

Allows auto-replacement of SQL's

```sql
SELECT CASE
    %%FIELD_MAPPING: class%%
    ELSE NULL
END, ...
```

with

```sql
SELECT CASE
    WHEN "subclass" IN ('school', 'kindergarten') THEN 'school'
    WHEN "shop"='bar'
        OR "subclass" IN ('alcohol','beverages','wine')
        THEN 'alcohol_place'
    ELSE NULL
END, ...
```

This PR is required for https://github.com/openmaptiles/openmaptiles/pull/693/files

### DIFF
--- a/README.md
+++ b/README.md
@@ -97,7 +97,7 @@ layer:
           subclass: ['school','kindergarten']
         alcohol_place:
           shop: ['bar']
-          subclass: ['alcohol','beverages','wine']
+          subclass: ['alcohol','beverages','wine%']
 schema:
   - ./building.sql
 datasources:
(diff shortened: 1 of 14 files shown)


eyusupov:
I'm going to rebase and test it more, seems much slower than before

nyurik:
> I'm going to rebase and test it more, seems much slower than before

Very strange - the generated SQL should be nearly identical to what we had before. This is just a pre-processor step... Let me know if you notice anything.

eyusupov:
sorry, I commented on a wrong PR.

PR #70, opened 2019-10-28
Stop DB connection on client gettile abort

nyurik:
If the browser cancels get tile request,
terminate database connection to avoid lengthy
unneeded queries running on the server.


nyurik:
CC: @eyusupov

PR #72, opened 2019-10-29
MVT query performance analysis tool

nyurik:
Use `test-perf` to evaluate tile generation performance against a PostgreSQL database.
This utility can test individual layers, several layers at once, as well as the whole tile.
It has several pre-defined testing areas to provide cross-test consistency.
Results are printed as histogram graphs, showing tile size distribution.
If multiple zoom levels are tested, another histogram shows per-zoom size distribution.
Run with `--help` to see all options.

image


klokan:
Looks really cool on the first sight - but I did not have time yet to practically try it...
Thanks @nyurik !

PR #83, opened 2019-11-04
Support for --gzip -- Postgres-side mvt gzipping

nyurik:
GZIP() is the functionality available as a new PostgreSQL extension,
written specifically for this use case by Paul Ramsey yesterday.

All hail Paul. https://github.com/pramsey/pgsql-gzip


klokan:
Wow! Thank you @pramsey!

(and of course, thanks @nyurik too)

PR #100, opened 2019-11-16
Import-sql: Remove pgclimb that break the build (https://github.com/urfave/cli upgrade)

francois2metz:
Pgclimb doesn't seems to be used, so removing it is the easy option.


nyurik:
@lukasmartinelli is there a reason pgclimb is not packaged in its own docker?
@eva-j / @klokan /  @daliborjanak do you use pgclimb in any of your internal processing, or can it be removed?   if pgclimb can be made into its own docker, we could remove it here even if it is used by you internally.

PR #101, opened 2019-11-18
Allow imposm of import_osm.sh to be configured

frodrigo:
Configuring imposm also on import allows to produce a right last.state.txt directly at import.

Report of https://github.com/openmaptiles/import-osm/pull/15

### DIFF
--- a/docker/import-osm/import_osm.sh
+++ b/docker/import-osm/import_osm.sh
@@ -25,7 +25,8 @@ function import_pbf() {
         -cachedir "$IMPOSM_CACHE_DIR" \
         -read "$pbf_file" \
         -deployproduction \
-        -write $diff_flag
+        -write $diff_flag \
+        -config "$CONFIG_JSON"
 }
 
 function import_osm_with_first_pbf() {


docker/import-osm/import_osm.sh
```diff
@@ -25,7 +25,8 @@ function import_pbf() {
         -cachedir "$IMPOSM_CACHE_DIR" \
         -read "$pbf_file" \
         -deployproduction \
-        -write $diff_flag
+        -write $diff_flag \
```
nyurik: what if `$diff_flag` or `$CONFIG_JSON` are not defined?  Wouldn't that break all existing users?  You might want to use this variant instead:
```suggestion
        ${diff_flag:+-write "$diff_flag"} \
```

docker/import-osm/import_osm.sh
```diff
@@ -25,7 +25,8 @@ function import_pbf() {
...
+        -config "$CONFIG_JSON"
```
nyurik:
```suggestion
        ${CONFIG_JSON:+-config "$CONFIG_JSON"}
```

docker/import-osm/import_osm.sh
frodrigo: It is computed here. But I does not touch this part except the trailing `\`.

https://github.com/openmaptiles/openmaptiles-tools/blob/9068be14e38e942a8b180b868c66e6d34cb8fb9a/docker/import-osm/import_osm.sh#L13-L18

docker/import-osm/import_osm.sh
frodrigo: I do the same as other imports commands.

https://github.com/openmaptiles/openmaptiles-tools/blob/9068be14e38e942a8b180b868c66e6d34cb8fb9a/docker/import-osm/import_osm.sh#L29

docker/import-osm/import_osm.sh
nyurik: sorry, my mistake about the -write flag, if it was there before, no need to change that

docker/import-osm/import_osm.sh
nyurik: @frodrigo I am not sure i understand what other commands you mean.  If there are users of `import_osm.sh`, adding a new argument with a new environment variable will break their code.
We wouldn't want that to happen - so to make it backwards compatible, we should handle
the case when `$CONFIG_JSON` is not set.

frodrigo: The default value are set in `Dockerfile` at https://github.com/openmaptiles/openmaptiles-tools/blob/9068be14e38e942a8b180b868c66e6d34cb8fb9a/docker/import-osm/Dockerfile#L51

PR #106, opened 2019-11-19
Update to the latest version of imposm.

francois2metz:
golang 1.9 is required with this version, so I updated the base image. It removes the use of the backports.

Also the resulting binary is now named `imposm`.


nyurik:
@francois2metz could you rebase on master - i have updated it to the latest golang base, and also did a hack so that both `imposm` and `imposm3` can both be used - they point to the same program.  I will still need to do a lot of testing to ensure that the new version can be used without breaking. Imposm did make a few changes. Plus there is a very promising pull request https://github.com/omniscale/imposm3/pull/211 that i suspect we may want to apply, thus continuing to fork until merged.

francois2metz:
It seems that you did all the job and his PR is not needed anymore.

nyurik:
> It seems that you did all the job and this PR is not needed anymore.

Sorry, didn't realize that the only thing left to do was to change default args.  I rebased. We may still want to apply the pending PR somehow.

PR #122, opened 2019-11-28
Parallelize import-water and import-osmborder

remi-dupre:
Parallelize the simplifications of water and osmborders. This can mainly help a lot for the dev workflow as these imports may be slow compared to other operations over a small region.

As this can lead to logs harder to interpret, this behavior is disabled by default.

This conflicts with the refactor of `import-water` (https://github.com/openmaptiles/openmaptiles-tools/pull/115). As I wasn't sure how to adapt I thought it would be better to just suggest this small change and get any advice on how to integrate it if needed.

### DIFF
--- a/docker/import-osmborder/README.md
+++ b/docker/import-osmborder/README.md
@@ -11,6 +11,7 @@ The **import-osmborder** will import the embedded CSV into the database.
 
 ```bash
 docker run --rm \
+    -e PARALLEL=1 \
     -e PGHOST="127.0.0.1" \
     -e PGDATABASE="openmaptiles" \
     -e PGUSER="openmaptiles" \
--- a/docker/import-osmborder/import_osmborder_lines.sh
+++ b/docker/import-osmborder/import_osmborder_lines.sh
@@ -62,45 +62,31 @@ function import_borders() {
     create_import_table "$table_name"
     import_csv "$csv_file" "$table_name"
 
-    local gen1_table_name="osm_border_linestring_gen1"
-    drop_table "$gen1_table_name"
-    generalize_border "$gen1_table_name" "$table_name" 10 10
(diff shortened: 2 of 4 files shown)


eva-j:
Thanks for this PR. We will test it.
Anyway, I suggest we will not merge it now before release of OMT 3.11 comes (December).

nyurik:
@eva-j would it make sense to just create a stable `v3.11` branch?  From the current master?  This way 3.11 flow won't be broken by any new merges, but you can always backport a patch to 3.11 if needed?

nyurik:
@remi-dupre thanks for your PR!  I do believe we should make paralelization better than what we have now, but I think there is a better way.  Your PR paralelizes just water, but it would be better to parallelize all data layers at once. I recently wrote a new tool [refresh-views](https://github.com/openmaptiles/openmaptiles-tools/blob/master/bin/refresh-views) which allows to refresh all parallel materialized view at the same time, taking into account view dependencies.

I think we should make import-* dockers simpler -- just import raw data into Postgres, and move all materialized views to the main OpenMapTiles repo as part of layer configuration.  Layers should use `WITH NO DATA` parameter in `CREATE MATERIALIZED VIEW`, making view creation very fast. After all views are added, we just run the `refresh-views`.  What do you think?

remi-dupre:
@nyurik, it indeed sounds much cleaner than how it's currently done !

Should I close this PR or should we still merge this rather small change even though it will no longer be required in some time ?

nyurik:
@remi-dupre i would rather just move the logic to the main tools repo after 3.11 is released, this way we don't need to test it and then retest it again.

nyurik:
Closing this PR as we will take a slightly different approach, per comments above.

PR #123, opened 2019-11-29
Adding admin_level 5 and 6 into _gen6 and _gen7 tables

eva-j:
Connected to https://github.com/openmaptiles/openmaptiles/pull/617.
I propose we accept this change even in the case https://github.com/openmaptiles/openmaptiles/pull/617 will not be merged - for more variability.

In future we should move this generalization step into https://github.com/openmaptiles/openmaptiles/blob/master/layers/boundary/boundary.sql.

### DIFF
--- a/docker/import-osmborder/import_osmborder_lines.sh
+++ b/docker/import-osmborder/import_osmborder_lines.sh
@@ -84,11 +84,11 @@ function import_borders() {
 
     local gen6_table_name="osm_border_linestring_gen6"
     drop_table "$gen6_table_name"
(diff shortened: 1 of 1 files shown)


nyurik:
@eva-j I think this is very similar to the issue in #122 - we don't want to bring map logic into the tools repo, but instead move all such rules into the main OMT repo.  The tools will contain border data, but it will be up to OMT layers to decide how to import them.

eva-j:
OK, make sense to wait with this till the generalization is in OMT.

PR #131, opened 2019-12-09
Allow to customize the zoom level used to expire tiles.

francois2metz:
This change is backward compatible. Useful when you need to expire a special zoom level.


nyurik:
thx!

Issue #132, opened 2019-12-10
Invalid psql version for import-osm docker


MartinMikita:
There is different PSQL version inside `import-osm` docker image (3.0.0), which does not correspond with `postgis` and `import-sql`.

I am not sure, whether it is used and required by `imposm` binary, but it should be the same.

 - [postgis uses PG_MAJOR=9.6](https://github.com/openmaptiles/openmaptiles-tools/blob/master/docker/postgis/Dockerfile#L1)
 - [import-sql uses PG_MAJOR=9.6](https://github.com/openmaptiles/openmaptiles-tools/blob/master/docker/import-sql/Dockerfile#L1)
 - [import-osm uses PG_MAJOR=11](https://github.com/openmaptiles/openmaptiles-tools/blob/master/docker/import-osm/Dockerfile#L4)

nyurik:
@MartinMikita, have you seen anything broken?  If not, this is not a bug. This is similar to how you would use the latest PostgreSQL client library in your code even if your server is old -- simply because other users may want to use a more recent server, and this way your client will support them all.

 `psql` has been mostly maintained to keep backwards compatibility. So while officially it is not required to do so, it has been working very well. On the other hand, connecting psql to a newer version of the server may create unexpected problems.

Some users (myself included) use different PostgreSQL versions, so unless something is broken (which doesn't sound like it is), I would prefer to keep the latest version in the client docker images.

The only reason `import-sql` is still 9.6 is because it is no longer needed.  All the code from import-sql is now part of the tools itself.  I will commit an updated version of the quickstart / docker-compose tomorrow showing how it can be used.

nyurik:
P.S.  See https://github.com/openmaptiles/openmaptiles-tools/blob/master/Dockerfile#L13 -- PG_MAJOR=12

MartinMikita:
Thanks for the feedback, it makes sense. I haven't seen any error, only warning, and differences.

In that case, why import-osm does have PG_MAJOR=11 when omt-tools has PG_MAJOR=12?

You should keep that in sync, otherwise, it could be broken.

nyurik:
I think it should be PG=12 everywhere. Moreover, I think we eventually should remove import-osm docker image, and just move the scripts to tools.

MartinMikita:
Thanks for your note. This ticket is not related anymore, the PSQL version was united with v4 release.

PR #133, opened 2019-12-10
Changed PG_MAJOR for import-osm (#132)

MartinMikita:
Solves #132 to corresponds with postgis and import-sql docker images


nyurik:
See my comments in #132

MartinMikita:
Will not be used, see @nyurik and #132

PR #139, opened 2019-12-13
Delete obsolete import-wikidata, import-sql

nyurik:
The import-wikidata and import-sql are no longer maintained,
and should not be used.  Docker hub will continue to host them,
but there is no point to have duplicate code in the repo.

This can be safely merged because OMT will continue using
dockerhub version until it is too is migrated.


nyurik:
I disabled the autobuild for docker hub `import-wikidata` and `import-sql` images.

PR #143, opened 2019-12-15
Rewrite tileset/layer classes with properties

nyurik:
Greatly simplifies usage of the parsed data,
and puts all parsing logic into one place.

More work will need to be done to migrate other usages.


klokan:
This is much cleaner now - in general instead of having `document['layer']['layer']` and different tests all around - the validity happens on a centralized place on getters/setters of attributes on the tileset data model.

+1

PR #144, opened 2019-12-17
+download-osm tool to get planet from all mirrors

nyurik:
Use `download-osm` tool to download the entire planet file (~50GB+) from all available OSM mirrors
at the same time. This distributes the load across the mirrors, and makes your download faster.
The tool will ensure you get the latest version, verifies that all mirrors contain the same
version, and validates the download with md5 hash. By default the tool will not download from
the primary OSM site to reduce its load.  Downloader uses [aria2c](https://aria2.github.io/).

```
# downloads the latest version to the current directory
download-osm planet

# download to the target dir by passing -d ... param to aria2c
download-osm planet -- -d ./downloads
```


bin/download-osm
```diff
@@ -214,62 +232,119 @@ class Site:
...
-    aria2c_args = args['<arguments>']
-
-    urls, md5 = await Catalog().init(verbose, use_primary)
-
-    params = ['aria2c', f'--checksum=md5={md5}']
+    dry_run = args['--dry-run']
+    aria2c_args = args.arguments
+
+    async with aiohttp.ClientSession(headers={
+        'User-Agent': 'OpenMapTiles download-osm '
```
MartinMikita: I would add version of OMT tools into User Agent

nyurik:
thanks @MartinMikita , added version.

PR #145, opened 2019-12-17
Customize postserve tile url and use the tileset for the metadata

francois2metz:
This add a new argument `--url` to be able format an url with the protocol and the hostname (without port).

Also postserve now use the tileset to fill the metadata instead of using hardcoded values. This remove the metadata set via environnement variables and breaks the compatibility. I can split this part if you want.


nyurik:
@francois2metz could you rebase this on top of #143 , which will for sure go in before this?  It will make tileset usage easier.  WRT the PR itself, we would have two semi-identical parameters - serve and url, which is very confusing.  If we are breaking backwards compatibility (no longer supporting ENV variables), we might as well also sligthly adjust the current --serve parameter. Or at least we could add some logic:
* if serve is None: url = 'localhost:port'
* if serve does not contains `:` -- url = serve + port
* if serve ends in `:80` -- url = serve without `:80`
* else url = serve

francois2metz:
We may try to reuse the serve parameter, by adding a `--protocol` parameter to configure http/https. What do you think ?

nyurik:
why not allow serve to start with http/https ?

francois2metz:
why not ;).Last thing, I want to configure the url in the metadata as `https://mytilesserver.com`, use the port `8090` (to not have a privilegied port) and have a reverse proxy in front of it. How to handle that?

nyurik:
i think we should simplify the whole thing -- if `serve` is given, use that full string without modifications. And port will not affect it.  If serve is not given, use `http://localhost:{port}` unless port is 80.

francois2metz:
Ok, it makes sense. I'll implement theses changes.

francois2metz:
I updated the `serve` parameter with a default value of `http://localhost:8090`. This should do the job. 
I'll rebase master as soon as your PR is merged.

nyurik:
thanks @francois2metz , merged, let me know

francois2metz:
Rebased

bin/postserve
```diff
@@ -17,7 +17,7 @@ Usage:
 Options:
   -l --layer=<layer>    If set, limit tile generation to just this layer (could be multiple)
   -x --exclude-layers   If set, uses all layers except the ones listed with -l (-l is required)
-  -s --serve=<port>     Return this hostname as tileserver URL in metadata  [default: localhost]
+  -s --serve=<url>      Return this URL as tileserver URL in metadata  [default: http://localhost:8090]
```
nyurik: Need to substitute default port for the `--port` param. Since docopt auto-sets this value, you can simply do a 
`url = args['--serve'].replace('<port>', args['--port'])`
using this change in doc:

```suggestion
  -s --serve=<url>      Return this URL as tileserver URL in metadata  [default: http://localhost:<port>]
```

francois2metz: Good catch. done.

PR #157, opened 2020-01-22
Numeric function and tests

jsanz:
**NOTE**: do not merge until #155 

Adds a new function to replace [`as_numeric`](https://github.com/openmaptiles/openmaptiles/blob/master/layers/building/building.sql#L4-L14) using the implementation from `postgis-vt-util`, adding tests to ensure consistency based in #155 testing framework.

Because of permissions I had to download the `postgis-vt-util` script into the `sql` folder, adding a cleanup line in the `Makefile` and `.gitignore`. I hope this is clean enough.


test-sql/numeric.sql
```diff
@@ -0,0 +1,11 @@
+SELECT omt_as_numeric(null) AS fail;
```
nyurik: I think it would be better to give each alias a unique name - this way it is easier to match the output file with each query.  E.g. here it could be `AS fail_null`

jsanz: yep, I initially did that in fact, agree that would help when errors raise.

jsanz:
> i just noticed that it fails CI due to rm *pre_tests.sql. Also, should we get the vt utils swtiched to the OMT-version first?

It's not necessary, both Mapbox and OMT `CleanNumeric` functions yield the same results even having a different implementation. In fact, this test can be used to check that just swapping the current Makefile line.

https://github.com/openmaptiles/openmaptiles-tools/blob/1262b8eca1c9dfc5cfa5d3d3099acbe6fb89f8e7/Makefile#L6

We can change the line to get the OMT version when we have #156 merged.

jsanz:
oooops sorry for bringing those commits from master here :sweat: @nyurik please let me know if it's better to start a new PR.

PR #158, opened 2020-01-23
Minor fix new "columns" imposm3 table def

nyurik:
We renamed "fields" to "columns" for imposm3 mapping definition,
but I forgot to update the import-wikidata script to detect those
too. Added the missing detection to support both, and up to 4.0.0


nyurik:
self-merging because its a minor issue that affects production pull requests to OMT

Issue #179, opened 2020-02-21
generate-imposm3, append tags > include


frodrigo:
If a layer define 
```yaml
tags:
    include: [...]
```
It override the default. It should append the included tags.

https://github.com/openmaptiles/openmaptiles-tools/blob/master/openmaptiles/imposm.py#L58

nyurik:
@frodrigo could you give a complete example of a layer with the tags? E.g. a PR to modify https://github.com/openmaptiles/openmaptiles-tools/blob/master/tests/testlayers/housenumber/mapping.yaml ? And also how that would change https://github.com/openmaptiles/openmaptiles-tools/blob/master/tests/expected/imposm3.yaml ?  Thanks!

PR #180, opened 2020-02-21
[WIP] Merge impsom tags include #179

frodrigo:
If a layer define
```yaml
tags:
    include: [...]
```
Now it override the default. It should in place append to the included tags.

### DIFF
--- a/openmaptiles/imposm.py
+++ b/openmaptiles/imposm.py
@@ -25,6 +25,7 @@ def create_imposm3_mapping(tileset_filename):
     include_tags.append('name')
     include_tags.append('wikidata')
     include_tags.append('wikipedia')
+    include_tags = set(include_tags)
 
     generalized_tables = {}
     tables = {}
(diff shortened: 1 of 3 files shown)


openmaptiles/imposm.py
```diff
@@ -55,8 +56,12 @@ def create_imposm3_mapping(tileset_filename):
                 definition.pop('_resolve_wikidata', None)
                 tables[table_name] = definition
             for tag_name, definition in mapping.get('tags', {}).items():
+                if tag_name == 'include':
+                    tags[tag_name] |= set(definition)
                 tags[tag_name] = definition
```
nyurik: you are overriding the above right after adding to it - should be an `else` there.  Also, you may want to add a type check (e.g. make sure `include` is a list of strings). 

In general, I think in that for loop we should be explicit - only allow the `include`, but prohibit everything else (raise an error)

Poke me on Slack, or I can update your PR in a bit

frodrigo: Changed

frodrigo:
Replaced by #181

PR #188, opened 2020-03-08
v5.0 Migrate import-osm => tools

nyurik:
### tools image
* Removed [pgfutter](https://github.com/lukasmartinelli/pgfutter) tool -- that tool's functionality was replaced with `psql -c '\copy ... CSV'` command (see import-data image)
* Image was refactored to be much smaller, and only contain build results
* removed /tileset and /sql volume declarations (no point in forcing those dirs to be volumes by default) -- they can still be mounted as external volumes.

### merged import-osm into tools
* All import-osm image tools were moved to the main tools repo -- `import-osm`, `update-osm`, and `import-diff` are now executed as any other tools from the main tools docker image.


TomPohys:
Thank you @nyurik for this PR. @MartinMikita and I need to check and verify the new tool v5 and OMT v5. Testing on the whole planet could take a week. We are working on it.

MartinMikita:
Hi @nyurik,
Is it possible to use `python:3.8-slim` as base image? We were able to build image, with small changes, and the resulted docker image is smaller - 760 MB instead of 1.36GB

```
diff --git a/Dockerfile b/Dockerfile
index 79df183..30bb227 100644
--- a/Dockerfile
+++ b/Dockerfile
@@ -55,7 +55,7 @@ RUN set -eux ;\
 
 
 # Primary image
-FROM python:3.8
+FROM python:3.8-slim
 ARG PG_MAJOR=12
 ARG VT_UTIL_VERSION=v2.0.0
 ARG TOOLS_DIR=/usr/src/app
@@ -80,7 +80,7 @@ ENV VT_UTIL_DIR=/opt/postgis-vt-util \
     CONFIG_JSON=${TOOLS_DIR}/config/repl_config.json
 
 
-
+RUN apt-get update -qq && apt-get install -y curl gnupg2
 RUN set -eux ;\
     /bin/bash -c 'echo ""; echo ""; echo "##### Installing packages..."' >&2 ;\
     curl https://www.postgresql.org/media/keys/ACCC4CF8.asc | apt-key add - ;\
```

Or do you have some notes, that it could break something? We would test this smaller image further.

nyurik:
Switched to use python-slim

MartinMikita:
Hi @nyurik,

You have created this PR 12 days ago and requested our review. We have started to test the whole process (this PR and PR in [OMT repository]()) at the beginning of the week.

We have reported a few problems, which should be repaired, and they were repaired on your side.
But you have introduced other changes (more or less functional like comments), not related to these problems - for example renaming env variables for dirs, changing import-oms/import-sql/importdiff/import-update binary scripts, updating dependency libraries.

This should happen before requesting our review and after your internal testing.
Now, we need to start the whole process over again, and we lost 4 days of our resources (time and money) to verify your PR.

bin/import-osm
```diff
@@ -20,36 +20,43 @@ function import_pbf() {
     local pbf_file="$1"
     local diff_flag=""
     if [ "$DIFF_MODE" = true ]; then
-        diff_flag="-diff"
```
MartinMikita: is this change necessary? Why is better to use the construction `${diff_flag:+"-diff"}`, instead of `$diff_flag` for the final command?

bin/import-diff
```diff
@@ -1,6 +1,5 @@
-#!/usr/bin/env bash
+#!/usr/bin/env sh
```
MartinMikita: Why are you changing `bash` into `sh`? Other `import-*` scripts uses `bash`, wouldn't be better to unify this, instead of making differences?

nyurik:
@MartinMikita appologies, I will revert and create a new PR, thus you won't have to retest

MartinMikita:
Related to the comment https://github.com/openmaptiles/openmaptiles/pull/785#issuecomment-601567895

It seems, that `import-sql` should [sort parallel SQL](https://github.com/openmaptiles/openmaptiles-tools/blob/v5.0-dev/bin/import-sql#L50), which is now missing. `find` utility is not sorting by default.

MartinMikita:
@nyurik Are you aware of any change, that could totally decrease the performance of `import-sql` step?
Using OMT tools `3.1.0` and OMT scheme `3.11` the import--

Dockerfile
```diff
@@ -1,74 +1,142 @@
...
+        ca-certificates \
+        cmake \
+        git \
+        libosmium2-dev \
+        zlib1g-dev \
+        ;\
+    /bin/bash -c 'echo ""; echo ""; echo "##### Building osmborder -- https://github.com/pnorman/osmborder"' >&2 ;\
+    git clone https://github.com/pnorman/osmborder.git /usr/src/osmborder ;\
+    cd /usr/src/osmborder ;\
+    git checkout ${OSMBORDER_REV?} ;\
```
MartinMikita: @nyurik can you explain, why there is sometimes construction like `${VAR?}` and sometimes like `${VAR:?}`?

Dockerfile
```diff
@@ -1,74 +1,142 @@
...
+    DEBIAN_FRONTEND=noninteractive apt-get update ;\
+    DEBIAN_FRONTEND=noninteractive apt-get install  -y --no-install-recommends \
+        # a few common tools
+        ca-certificates \
+        curl \
+        wget \
+        git  \
+        gnupg2  `# TODO: not sure why gnupg2 is needed`  ;\
+    curl https://www.postgresql.org/media/keys/ACCC4CF8.asc | apt-key add - ;\
+    /bin/bash -c 'source /etc/os-release && echo "deb http://apt.postgresql.org/pub/repos/apt/ ${VERSION_CODENAME:?}-pgdg main ${PG_MAJOR:?}" > /etc/apt/sources.list.d/pgdg.list' ;\
```
MartinMikita: Here `${VAR:?}`

Dockerfile
```diff
@@ -1,74 +1,142 @@
...
-        `# tools to build osmborder` \
-        build-essential \
-        ca-certificates \
-        cmake \
-        git \
-        libosmium2-dev \
-        wget \
-        zlib1g-dev \
+        osmosis   `# useful toolset - https://wiki.openstreetmap.org/wiki/Osmosis` \
+        postgresql-client-${PG_MAJOR:?}  `# psql` \
```
MartinMikita: Here `${VAR:?}`

Dockerfile
```diff
@@ -1,74 +1,142 @@
...
+    /bin/bash -c 'echo ""; echo ""; echo "##### Cleaning up"' >&2 ;\
+    rm -rf /var/lib/apt/lists/*
 
 # Copy requirements.txt first to avoid pip install on every code change
 COPY ./requirements.txt .
-RUN pip install --no-cache-dir -r requirements.txt
 
+RUN set -eux ;\
+    pip install --no-cache-dir -r requirements.txt ;\
+    mkdir -p "${VT_UTIL_DIR?}" ;\
```
MartinMikita: But here `${VAR?}`

Dockerfile
MartinMikita: In the main and current release, it is [`${VAR?}`](https://github.com/openmaptiles/openmaptiles-tools/blob/master/Dockerfile#L27)

nyurik:
WRT `${VAR?}` vs `${VAR:?}` -- this is almost the same, and we should eventually use `:?` everywhere. The `${VAR?}` would raise an error if VAR is not defined, whereas `:?` would raise an error if VAR does not exist or is empty.

WRT slowdown:  the only big change is the version of PostgreSQL db.  postgis image was based on 9.6, and now it is based on PG11.

I did not run full planet import on postgis in docker (as several Postgres engineers recommended not using docker for real database) using a  n1-highmem-16 GCP instance (16 CPUs, 104GB RAM) -- installing latest `Debian sid` versions of Postgres 12 + Postgis 3.0.1.    The importer (imposm and other tools) were running in Kubernetes on another n1-highmem-16 machine, but this was clearly on overkill.  I accidentally forgot to optimize PostgreSQL settings before the run, so my results should be significantly worse than would could have been achieved with optimization.
* imposm reading step - 1h5m, writing 11h9m, fully done - 15h23m
* import-sql - 10h
* analyze - 8m

<details><summary>Settings I plan to use (expand)</summary>

```
# 25% of the total memory
shared_buffers=$(awk '/MemTotal/ { printf "%d", $2/1024 * 0.25 }' /proc/meminfo)

  maintenance_work_mem = 256MB
  work_mem = 128MB
  max_locks_per_transaction = 128

  min_wal_size = 256MB
  max_wal_size = 50GB
  wal_keep_segments = 128
  wal_sender_timeout = 300s
  wal_level = replica
  wal_log_hints = on

  hot_standby = on
  archive_mode = off
  hot_standby_feedback = on

  max_wal_senders = 90
  max_connections = 120
  max_replication_slots = 50

  checkpoint_timeout = 15min
  checkpoint_completion_target = 0.8
  random_page_cost = 1.0
```
</details>

bin/import-osm
nyurik: This is not required, but it is a suggested solution for [SC2086](https://github.com/koalaman/shellcheck/wiki/SC2086) -- look for `--trace-commands` example there.

bin/import-diff
nyurik: If we don't any extra features of bash, we may want to eventually switch to alpine base, which doesn't have bash (currently the `import-data` image is the only alpine-based one). No biggie here, up to you.

TomPohys:
Hi @nyurik, thank you for this PR with a lot of enhancements.

 After extensive internal testing of tools v5, there is a significant performance decrease in generating tiles between tools v5 and v4.1. For generating areas with more than 1k tiles, tools v5 prolongs tile generating by almost 30%. 

Can be PR #188 without any change of PostgreSQL and PostGIS and keep this PR with the migration of `import-osm` into tools, merge "static" import data into `import-data` and preloaded-PostGIS for testing?

Update to new version of PostgreSQL and PostGIS can be in separate PR.

Or do you have other suggestion about how to handle this?

nyurik:
@TomPohys per your suggestion, I reverted to 9.6 postgres & old imposm

PR #191, opened 2020-03-09
Implement Borders & Wikidata local caching

### LINKED ISSUE #194: Make borders and wikidata cachable and able to import from backup
* We should be able to import current (new) borders from PBF, but also import from backup CSV 
* The same principle with wikidata - Use Wikidata Quering Service for updates and new queries, but also able to import pre-cached, backup data. 

Borders are in CSV, wikidata maybe in SQLite

nyurik:
Written in collaboration with Tom Pohanka <tompohys@gmail.com>

* Wikidata labels can now be cached in a local file, and imported from that file as needed
* border importer can now be ran to just import the borders csv file, without PBF parsing

Fixes #194


bin/import-wikidata
```diff
@@ -87,7 +87,7 @@ async def main(args):
     if cache:
         # Do not query for the IDs we already have in cache
         new_ids = set(ids).difference(cache.keys())
-        print(f'Found {len(ids):,} Wikidata IDs, {len(new_ids) - len(ids):,} in cache, '
+        print(f'Found {len(ids):,} Wikidata IDs, {abs(len(new_ids) - len(ids)):,} in cache, '
```
nyurik: that's not a noop -- it was a bug and `abs()` hides it :).  I have fixed it by changing the order of what is subtracted from what.

PR #207, opened 2020-03-26
Order MVT layers in the output, debug-mvt summary

nyurik:
With the introduction of parallel query execution, the order of data
became less predictable, thus making tiles change their hash even when
no changes were made.  This patch attempts to solve it at least for
the layer ordering -- each layer is given an index, and the layers
are sorted right after the `UNION ALL` clause.

Additionally, debug-mvt tool has gained an ability to get the summary,
tile size (compressed and uncompressed), key, and time it took.


nyurik:
Closing in favor of #232

PR #216, opened 2020-04-13
`osmborder` support for openstreetmap.fr extracts

### LINKED ISSUE #215: `osmborder` aborts on openstreetmap.fr extracts
`osmborder` aborts when its input is an extract downloaded from [openstreetmap.fr](http://download.openstreetmap.fr/extracts/)

### Error Message
```
terminate called after throwing an instance of 'osmium::not_found'
  what():  location for one or more nodes not found in node location index
```

### Analysis
Note: 14-Apr-20 edits are marked **in bold below**
- The error message suggests that the input file contains references to nodes that are located outside of the extraction polygon.
   ```
   $ docker-compose run --rm openmaptiles-tools osmborder /import/borders/filtered.pbf -o /dev/null
   terminate called after throwing an instance of 'osmium::not_found'
     what():  location for one or more nodes not found in node location index
   ```
- `osmborder` succeeds if **before running `osmborder_filter`** the input file is fixed using  
**`osmconvert --drop-broken-refs -b=-180,-90,180,90`**.</br>
   <b>
   ```
   $ docker-compose run --rm openmaptiles-tools osmconvert --drop-broken-refs -b=-180,-90,180,90 -o=/import/borders/monaco.osm.pbf /import/monaco.osm.pbf

   $ docker-compose run --rm openmaptiles-tools osmborder_filter -o /import/borders/filtered.pbf /import/borders/monaco.osm.pbf

   $ docker-compose run --rm openmaptiles-tools osmborder -o /import/borders/lines.csv /import/borders/filtered.pbf
   There were 0 warnings.
   There were 0 errors.

   $ docker-compose run --rm openmaptiles-tools wc -l /import/borders/lines.csv
   72 /import/borders/lines.csv
   ```
   </b>

### To Reproduce
In an [`openmaptiles`](https://github.com/openmaptiles/openmaptiles) installation run:
```
rm -f data/*.pbf
docker-compose run --rm openmaptiles-tools bash -c 'download-osm osmfr europe/vatican_city -- -d /import'
make import-borders
```

Output:
```
$ rm -fr data

$ docker-compose run --rm openmaptiles-tools bash -c 'download-osm osmfr europe/vatican_city -- -d /import'
Unable to load md5 hash for europe/vatican_city from http://download.openstreetmap.fr/extracts/europe/vatican_city-latest.osm.pbf: Received status=404
Downloading http://download.openstreetmap.fr/extracts/europe/vatican_city-latest.osm.pbf (size=0.5 MB (488,724), md5=None)

  aria2c --http-accept-gzip "--user-agent=OpenMapTiles download-osm 4.1.0 (https://github.com/openmaptiles/openmaptiles-tools)" -d /import http://download.openstreetmap.fr/extracts/europe/vatican_city-latest.osm.pbf


04/13 07:40:06 [NOTICE] Downloading 1 item(s)
[#49af17 304KiB/477KiB(63%) CN:1 DL:368KiB]
04/13 07:40:08 [NOTICE] Download complete: /import/vatican_city-latest.osm.pbf

Download Results:
gid   |stat|avg speed  |path/URI
======+====+===========+=======================================================
49af17|OK  |   468KiB/s|/import/vatican_city-latest.osm.pbf

Status Legend:
(OK):download completed.

$ make import-borders
docker-compose up -d postgres
openmaptiles_postgres_1 is up-to-date
Wait for PostgreSQL to start...
docker-compose run --rm import-osm ./pgwait.sh
docker-compose run --rm openmaptiles-tools import-borders
Filtering /import/vatican_city-latest.osm.pbf into /import/borders/filtered.pbf
Creating a CSV borders file /import/borders/lines.csv
terminate called after throwing an instance of 'osmium::not_found'
  what():  location for one or more nodes not found in node location index
/usr/src/app/import-borders: line 75:    23 Aborted                 osmborder -o "$BORDERS_CSV_FILE" "$BORDERS_PBF_FILE"
Makefile:158: recipe for target 'import-borders' failed
make: *** [import-borders] Error 134
```

#### Note
This new issue was opened in this repository because it uses a specific version of [osmborder](https://github.com/pnorman/osmborder)
   https://github.com/openmaptiles/openmaptiles-tools/blob/4ed919770276a1aad949c7ddfe4bbfa0b68e3196/Dockerfile#L12

zstadler:
openstreetmap.fr tends to generate extracts with the non-existing refs, and needs to be cleaned up.
This patch uses `osmconvert --drop-broken-refs ...` on the PBF file before generating borders file.

Resolve #215

Co-authored-by: Yuri Astrakhan <YuriAstrakhan@gmail.com>


zstadler:
Unfortunately, running `osmconvert --drop-broken-refs` 
https://github.com/openmaptiles/openmaptiles-tools/blob/620563414b7732ef74a69be4f0df3beb434f66d8/bin/import-borders#L65
was unable to remove the references to the missing nodes in the case of the monaco extract.

bin/import-borders
```diff
@@ -60,7 +60,10 @@ if [[ "$task" == "import" || "$task" == "parse" ]]; then
   echo "Filtering $IMPORT_PBF_FILE into $BORDERS_PBF_FILE"
   mkdir -p "$(dirname "${BORDERS_PBF_FILE?}")"
   rm -rf "${BORDERS_PBF_FILE?}"
-  osmborder_filter -o "$BORDERS_PBF_FILE" "$IMPORT_PBF_FILE"
+  rm -rf "${BORDERS_PBF_FILE%.pbf}-temp.pbf"
```
nyurik, line 63: Please make `"${BORDERS_PBF_FILE%.pbf}-temp.pbf"` into a var and use it here and below. Also, add a new `echo "Dropping broken references (file) -> (file)"`  (or something similar) -- we should indicate the extra step.  The `echo` from above could be moved right before the `osmborder_filter` call.

Also, I do not want this step to always execute, esp when dealing with full planet -- as that would take significant time and won't be much use.  We could introduce a new "task" (command) -- e.g. `import-fix` and `parse-fix` (name suggestions are welcome), or it could be done via an env var.  This way fixing the file would become an optional step.

Lastly, I think this is a temporary workaround, and it should be the `osmborder_filter` that should handle it automatically. Please create an upstream issue in the @pnorman 's https://github.com/pnorman/osmborder

BTW, I think the environment variable path (rather than adding new tasks) would be preferable, because once it is fixed upstream, we can simply ignore that env var.

bin/import-borders
```diff
@@ -57,14 +60,27 @@ task="$1"
 if [[ "$task" == "import" || "$task" == "parse" ]]; then
   # Parse PBF file to generate borders CSV file, and import it into PG
   IMPORT_PBF_FILE="$2"
-  echo "Filtering $IMPORT_PBF_FILE into $BORDERS_PBF_FILE"
+
+  mkdir -p "$(dirname "${BORDERS_CLEANUP_FILE?}")"
+  rm -rf "${BORDERS_CLEANUP_FILE}"
+  if [[ "${BORDERS_CLEANUP?}" == "true"]; then
```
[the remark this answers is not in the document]
zstadler: +1

bin/import-borders
```diff
@@ -6,6 +6,8 @@ set -o nounset
 : "${IMPORT_DIR:=/import}"
 # Do not use .pbf extension to avoid accidental conflict with the original pbf with multiple runs
 : "${BORDERS_PBF_FILE:="$IMPORT_DIR/borders/filtered.pbf"}"
+: "${BORDERS_CLEANUP:=true}"
+: "${BORDERS_CLEANUP_FILE:="$IMPORT_DIR/borders/cleanup.pbf"
```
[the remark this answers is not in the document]
zstadler: +1

bin/import-borders
```diff
@@ -4,10 +4,10 @@ set -o pipefail
 set -o nounset
 
 : "${IMPORT_DIR:=/import}"
-# Do not use .pbf extension to avoid accidental conflict with the original pbf with multiple runs
+: "${BORDERS_CLEANUP:=false}"
```
zstadler, line 7: I suggest the cleanup should be turned on by default.
I believe most users do not use Openmaptiles for large extracts such as planet. Moreover, those who do use it for planet are usually more skilled and would know the internals better.

bin/import-borders
```diff
@@ -4,10 +4,10 @@ set -o pipefail
...
+: "${BORDERS_CLEANUP_FILE:="$IMPORT_DIR/borders/cleanup.pbf"}"
+: "${BORDERS_CLEANUP_BBOX:=-180,-90,180,90}"
```
zstadler, line 9: the cleanup orders should not change as this is just a trigger to activate --drop-broken-refs. It is set to the maximum in order to avoid any extra computation in osmconvert.

bin/import-borders
nyurik: ok, agree to both changes - could you submit them and i will merge?  Assuming you are ok with the other changes i made?

zstadler: Done.

bin/import-borders
zstadler: Done

zstadler:
@nyurik,
BTW, it is not possible to create new issues for [pnorman/osmborder](https://github.com/pnorman/osmborder/issues):
> This repository has been archived by the owner. It is now read-only.

frodrigo:
If task force is required to maintain osmborder, I may be interested. I also need it and improvement on it.

nyurik:
Maybe @pnorman might be interested in transferring it to omt ?

zstadler:
Please see a proposal for an alternative approach at https://github.com/openmaptiles/openmaptiles-tools/issues/215#issuecomment-613870437

nyurik:
> Unfortunately, running `osmconvert --drop-broken-refs`
> https://github.com/openmaptiles/openmaptiles-tools/blob/620563414b7732ef74a69be4f0df3beb434f66d8/bin/import-borders#L65
> 
> 
> was unable to remove the references to the missing nodes in the case of the monaco extract.

@zstadler sorry I missed this comment. Are you saying this code will no longer work after merging in case users will use Monaco extract?

zstadler:
> 
> 
> > Unfortunately, running `osmconvert --drop-broken-refs`
> > https://github.com/openmaptiles/openmaptiles-tools/blob/620563414b7732ef74a69be4f0df3beb434f66d8/bin/import-borders#L65
> > 
> > was unable to remove the references to the missing nodes in the case of the monaco extract.
> 
> @zstadler sorry I missed this comment. Are you saying this code will no longer work after merging in case users will use Monaco extract?

That's the reason for the follow-up commit - https://github.com/openmaptiles/openmaptiles-tools/pull/216/commits/237fe325e4fd55de28ebfdcc126155316c439d17.

- The cleanup with `osmconvert` does not work unless is is done **before** running `osmborder_filter`.
- The `--drop-broken-refs` flag is not active unless `osmconvert` is also given one of its clipping flags. I chose to use `-b=-180,-90,180,90` in order to avoid any real clipping activity.

Otherwise, `osmborder` creates an empty table. Now that running the cleanup is the default, the monaco test shows that the table created by the cleanup flow is the same as the reference table.

TomPohys:
Thank you for PR. But if it affects just openstreetmap.fr we should make it opt-in to not clean borders every time (especially for large files). For Geofabrik it is not necessary.

@nyurik please merge it if OK with default is `false`.

PR #235, opened 2020-05-01
Upgrade to PG 11 + Postgis 3.0.1

nyurik:
* Use proper postgis-maintained docker image rather than build our own
* use UTF8PROC 2.5.0
* include postgres gzip support


nyurik:
Replaced by #270

PR #236, opened 2020-05-01
Upgrade to latest Imposm version

nyurik:
Use v0.10.0 instead of custom fork from v2017-10-18

Note that Imposm v0.10.0 is already being published as official `5.2-imposm10` docker images in addition to the default `5.2` ones.  This PR would stop publishing the additional `-imposm10` images and make it the default.


amatissart:
If it has not been addressed yet, this issue may need your attention before upgrading imposm:
https://github.com/openmaptiles/openmaptiles/issues/487

TomPohys:
Imposm >= 0.5.0 may import invalid geometries. from: https://github.com/openmaptiles/openmaptiles/issues/487

nyurik:
I have adjusted our build process so that `5.2.0-imposm10` is automatically published together with `5.2.0` (for all images). Will keep it this way until we are ready to merge v10 (or maybe start maintaining it ourselves?)

frodrigo:
Note the current of imposm version is now 0.11.1.

frodrigo:
I do not see any thing in the openmaptiles/imposm3 branch v2017-10-18 fork that is not on the current imposm3 master.
Why we are still stuck on the fork?
As I can see, the the fork is just an old version with no addition.

TomPohys (approved):

PR #245, opened 2020-05-05
remove hardcoded testcases before releases

nyurik:


nyurik:
Github is having issues with actions at the moment, need to push to force rebuild :(

PR #252, opened 2020-05-14
Upgrade to Postgres12, latest ogr2ogr in preload

nyurik:
* Use PostgreSQL 12
* skip any features with errors when importing NaturalEarth data
* Just for the preload image, use the latest ogr2ogr from `sid` (unstable)


Duiesel:
@nyurik Hi, Can't figure it out in what state is this PR now? Is there any problems with ogr2ogr and PG12? Or it will be merged later?

Duiesel:
@nyurik I've tried ogr2ogr 3.0.4 with `-skipfailures` and got at least one empty `ne_10m_ocean` table as a result. So this option is useless in this case. May be this is bug in the GDAL source.

TomPohys:
Could we close this when [#413](https://github.com/openmaptiles/openmaptiles-tools/pull/413) is already merged?

TomPohys:
Thanks! Let's close it with regard to https://github.com/openmaptiles/openmaptiles-tools/pull/413.

PR #255, opened 2020-05-16
download-osm improvements

nyurik:
* Added `--output <file>` param to specify the location and the name of the file to save. Also allows `--force` to override existing file.
* allow area and URL downloading without explicitly specifying the source:
```
download-osm toronto      # will search Geofabrik, BBBike, and osm.fr
download-osm https://...  # will download from URL
```
* added cacheable BBBike catalog (scraped from a single HTML page), allowing non-exact searches (`download-osm bbbike toronto` instead of `Toronto`) and catalog listing (`download-osm list bbbike`)
* better handling of mirrors that have just the latest planet file
* added a few mirrors

* **CHANGE:** `--make-dc` will now set `OSM_AREA_NAME` to the full ID of the area, e.g. for Geofabrik it will be `europe/monaco` instead of `monaco`


zstadler:
1. Since the source name is optional, I suggest it will be placed before the other optional arguments:
   ```
   download-osm <area_or_url> [url|geofabrik|osmfr|bbbike] [...] ...
   ```

2. Use `[-- [aria2c_args]]` to clarify the use of `--`. 
2. I suggest a separate flag, say `--refresh-catalog`, will be used for requesting a refresh of a source catalog, rather than reusing planet's `--force-latest` flag.

3. Previously, the  `--id=<area_id>` could be used only for `url` sources. I don't understand the how it would be interpreted for `geofabrik`, `osmfr`, or `bbbike`. 
   Please consider keeping the two syntax formats separate:
   ```
   download-osm <url> [url] [-o <file> [-f]] [-s <file>] [-c] [-n] [-v]
      [-d <file> [-i <z>] [-a <z>] [-x <v>] [--id <id>]] [[-- [aria2c_args]]
   download-osm <area> [geofabrik|osmfr|bbbike] [-o <file> [-f]] [-s <file>] [-c] [-n] [-v]
      [-j <file> [-k <kv>]...] [--refresh-catalog] 
      [-d <file> [-i <z>] [-a <z>] [-x <v>]] [-- [aria2c_args]]
   ```
   _note:_ I didn't see any use of the `OSM_AREA` value that `--make-dc` creates. Perhaps it could be removed!?

4. What is the logic used for "Area source has not been specified. Auto-detecting..."? Does it implement the logic implemented in https://github.com/openmaptiles/openmaptiles/pull/861?

nyurik:
> 1. Since the source name is optional, I suggest it will be placed before the other optional arguments:
>    ```
>    download-osm <area_or_url> [url|geofabrik|osmfr|bbbike] [...] ...
>    ```
I am not too sure about this -- it would break backward compatibility. The only benefit I can see is that it makes quickstart code a bit simpler (not even sure about that it really does).

> 2. Use `[-- [aria2c_args]]` to clarify the use of `--`.
thx, fixed

> 3. I suggest a separate flag, say `--refresh-catalog`, will be used for requesting a refresh of a source catalog, rather than reusing planet's `--force-latest` flag.


> 4. Previously, the  `--id=<area_id>` could be used only for `url` sources. I don't understand the how it would be interpreted for `geofabrik`, `osmfr`, or `bbbike`.
>    Please consider keeping the two syntax formats separate:
>    ```
>    download-osm <url> [url] [-o <file> [-f]] [-s <file>] [-c] [-n] [-v]
>       [-d <file> [-i <z>] [-a <z>] [-x <v>] [--id <id>]] [[-- [aria2c_args]]
>    download-osm <area> [geofabrik|osmfr|bbbike] [-o <file> [-f]] [-s <file>] [-c] [-n] [-v]
>       [-j <file> [-k <kv>]...] [--refresh-catalog] 
>       [-d <file> [-i <z>] [-a <z>] [-x <v>]] [-- [aria2c_args]]
>    ```

This one is tricky -- the docopt parser might not be able to handle two identical patterns that only differ in the url variable. I will see if i can change that.

>    
>    _note:_ I didn't see any use of the `OSM_AREA` value that `--make-dc` creates. Perhaps it could be removed!?
Possibly -- might be a different PR if agreed.

> 5. What is the logic used for "Area source has not been specified. Auto-detecting..."? Does it implement the logic implemented in [openmaptiles/openmaptiles#861](https://github.com/openmaptiles/openmaptiles/pull/861)?

No, the logic is as follows:
* if starts with `http://` or `https://` -- try it as a URL
* look through Geofabrik catalog, matching the last portion of the name (e.g. `michigan` would match `us/michigan`)  -- this is the same logic as before
* look through BBBike catalog (string-insensitive comparison, must be a full match)
* check if osm.fr gives a 200 rather than 404
* fail otherwise

PR #256, opened 2020-05-16
Clean Mapnik output and errors

zstadler:
- Remove "is deprecated" error messages
- Redraw progress on the same line


TomPohys:
Hi @zstadler, thanks for these changes. I am OK with them while they are optional.

zstadler:
I'm OK if they are the default

nyurik:
thanks! I just really hope we can stop using this antiquated system to generate tiles.

PR #296, opened 2020-06-10
Remove the double set of the header Access-Control-Allow-Origin.

francois2metz:
The header `Access-Control-Allow-Origin` is also set in the parent class `RequestHandledWithCors`.


nyurik:
thanks!

PR #297, opened 2020-06-11
Remove download-osm make-dc, gen bbox

nyurik:
Simplify `download-osm` to only generate bbox value, without the overly complex (and unneeded) docker-compose config file.

New functionality:
* `--bbox` will generate a simple text file containing the bounding box string. The value will be either computed based on the geofabrik's catalog, or for other sources it will continue using `osmconvert`
* `download-osm bbox <pbf-file> <bbox-file>` will parse pbf file to generate the same text file with bbox string
* The `--output` param is now required when used with `--bbox`

Removed functionality:
* `--make-dc` parameter and `make-dc` command.


TomPohys (approved):

TomPohys:
Thanks @nyurik, looks great!

zstadler:
Thanks @nyurik for simplifying the interface between `download-osm` and `openmaptiles`!

There are significant difference between the the polygon bbox of _openstreetmap.fr_ and the bbox computed by `osmconvert`.
For example, the actual bbox of the _israel_and_palestine_ polygon is `33.05,29.435,35.915,33.355` while the computed bbox is `25.1488889,22.0000374,36.8999418,33.4835140`. 
This is 18x increase in size that impacts the tile generation time.

openstreetmap.fr has its polygons at `https://download.openstreetmap.fr/polygons/$(area}.poly`.

Would it be possible to use them in `download-osm` for accurate bbox generation?

For example, https://download.openstreetmap.fr/polygons/europe/monaco.poly is
```
polygon
1
	7.406	43.732
	7.41	43.738
	7.419	43.743
	7.42	43.745
	7.425	43.746
	7.425	43.749
	7.429	43.753
	7.433	43.753
	7.436	43.756
	7.441	43.755
	7.443	43.752
	7.457	43.745
	7.537	43.538
	7.537	43.535
	7.535	43.533
	7.501	43.512
	7.497	43.514
	7.415	43.722
	7.408	43.725
	7.405	43.728
	7.406	43.732
END
END
```

For our purpose, we only care about the `  <lon>  <lat>` lines. All `END` lines as well as single number lines, like `1` and `2` should be ignored.

nyurik:
@zstadler sure, we could improve the bbox generation further. Let me know if you want to submit a PR for that.

zstadler:
@nyurik,

Unfortunately, python is not my forte...

zstadler:
> 
> 
> @nyurik,
> 
> Unfortunately, python is not my forte...

https://github.com/ustroetz/polygon2osm/blob/master/polygon2geojson.py converts a `.poly` file to geojson

nyurik:
@zstadler https://github.com/openmaptiles/openmaptiles-tools/pull/326

PR #301, opened 2020-07-02
Misnamed script in README.md

mattcen:


mattcen:
Worth noting that in https://github.com/openmaptiles/openmaptiles/blob/master/Makefile#L335, `update-osm` is the name of the Makefile target, but the actual script in this repo that it calls is named `import-update`

nyurik:
thanks!

PR #307, opened 2020-08-10
Use print_err in places that mix errors and output on stdout.

lseelenbinder:
We use these scripts in CI and since most of the scripts push results to `stdout`, we redirect the result to a file. This works great, until there's an error.

When there's an error, a few of the scripts push the errors to `stdout` instead of `stderr`. This means it appears the commands succeeded when they actually failed.

This commit adjusts some `print` statements to use `print_err` to ensure these results go to `stderr`.

### DIFF
--- a/bin/generate-sql
+++ b/bin/generate-sql
@@ -18,6 +18,7 @@ from pathlib import Path
 from docopt import docopt
 import openmaptiles
 from openmaptiles.sql import collect_sql
+from openmaptiles.utils import print_err
 
 if __name__ == '__main__':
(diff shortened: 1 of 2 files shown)


nyurik:
We use these scripts in CI and since most of the scripts push results to `stdout`, we redirect the result to a file. This works great, until there's an error.

When there's an error, a few of the scripts push the errors to `stdout` instead of `stderr`. This means it appears the commands succeeded when they actually failed.

This commit adjusts some `print` statements to use `print_err` to ensure these results go to `stderr`.

PR #310, opened 2020-08-20
Add generate-vectortiles-pg tool.

lazaa32:
Related to https://github.com/openmaptiles/openmaptiles/issues/875
New omt-tool `generate-vectortiles-pg` which used ST_AsMVT instead of Mapnik.


francois2metz:
@nyurik wrong mention ;)

nyurik:
Dope, autocomplete caught me :)

lazaa32:
@nyurik I use this tool in my fork of OMT and use it same way like the `generate-vectortiles`. That means I have a separate make target `make generate-tiles-pg` which use this docker. I just kept the same structure and flow. 
If I understand correctly then you propose to move this into `/bin` dir a use `openmaptiles-tools` image to run this tool? We can do that too, I am just confused why the original mapnik `generate-vectortiles` has a separate image (it could be move to bin folder as well, couldn't be?) Thanks for explanation.

nyurik:
@lazaa32 historical reasons :) See https://github.com/openmaptiles/openmaptiles/issues/595 for all the fun history.  The generate one is the only one that hasn't been migrated, and for a good reason -- I want to eventually get rid of it entirely, and the effort of migrating it to the `bin/`, with all the legacy nodejs + Mapnik is not worth it.  The MVT-based generation is much more straightforward - requires just the simple tilelive-copy + sqlite + postgres support, which is already part of the main docker.

lazaa32:
@nyurik Ok, I can move the `export-local-pgquery.sh` to `bin` and remove the docker folder `generate-vectortiles-pg`. Should we rename the bash script to something like `generate-tiles-pgquery.sh`?

nyurik:
Lets keep two phases separate:
* phase one -- introduce the new script, mark the old docker image as deprecated (possibly show a big warning when it runs?), and once merged and we release a new version, migrate the main OMT repo to use it.
* phase two -- once everyone is happy with the new approach, delete the old image. This might take some time.

Naming: let's keep it a bit shorter, perhaps `generate-tiles` for consistency with others?  Naming TBD :)

klokan:
Good points @nyurik 
We are in "phase one -- introduce the new script".
Let's merge and follow with other steps iteratively.

openmaptiles/__init__.py
```diff
@@ -1 +1 @@
-__version__ = "5.2.0"
+__version__ = "test-5.0-pg12-3.0"
```
nyurik, line 1: why is this change here?

nyurik:
There is still a strange change to the version file - is it really needed?  Also, lets shorten the script to `generate-tiles` (without the extension) -- makes it cleaner.

openmaptiles/__init__.py
lazaa32: Hi @nyurik, this PR is into `test-5.0-pg12-3.0` branch because it is the only one which has correct version of PostgreSQL and PostGIS. I didn't make the PR into master because it didn't make sense to me add it there when the script requires PostgreSQL12+. This `__version__` is used as a tag for a Docker image during the build  and I though the tag should correspond to branch name. However if that is not correct please let me know what version should be there and whether I chose a right branch. Thanks!

lazaa32:
Hi @nyurik, could you please let me know what is the right solution for version tag in the reviewed change above? Thanks!

openmaptiles/__init__.py
nyurik: @lazaa32 i think we shouldn't change it here because ST_AsMVT works in Postgres v9.6 just as well (technically it was first implemented in PostGIS 2.4, and we have been using it ever since). If you want to use PG12, just set the appropriate var in the .env file.

lazaa32:
Hi @nyurik , thanks for explanation. I will keep the same version then. I will make a new PR into `master` branch instead of current `v5.0-pg12-3.0`. 

> @lazaa32 historical reasons :) See [openmaptiles/openmaptiles#595]
The MVT-based generation is much more straightforward - requires just the simple tilelive-copy + sqlite + postgres support, which is already part of the main docker.

However I can't see tilelive-copy to be part of the main Docker image. Am I wrong?

nyurik:
> Hi @nyurik , thanks for explanation. I will keep the same version then. I will make a new PR into `master` branch instead of current `v5.0-pg12-3.0`.
> 
> > @lazaa32 historical reasons :) See [[openmaptiles/openmaptiles#595](https://github.com/openmaptiles/openmaptiles/issues/595)]
> > The MVT-based generation is much more straightforward - requires just the simple tilelive-copy + sqlite + postgres support, which is already part of the main docker.
> 
> However I can't see tilelive-copy to be part of the main Docker image. Am I wrong?

What do you mean you cannot see it?  Just add the new script to the `bin/` dir, and modify the main `Dockerfile` a bit similar to what I did in https://github.com/nyurik/tilelive-pgquery/blob/master/Dockerfile

lazaa32:
Closed. A new PR #317 into master was created.

PR #315, opened 2020-09-16
Include all *name* tags in `slice_language_tags`

zstadler:
When tags like `mtb:name:en` or `seamark:name` are included in `tags` mapping section, the existing implementation of `slice_language_tags` prevents their inclusion in the `transportation_name` and `waterway` layers.

This update adds to the result of `slice_language_tags` all `*name*` tags where the optional prefix and and suffix are separated by a `:` or a `_` character.

The added code is based on 
https://github.com/openmaptiles/openmaptiles-tools/blob/c310d1a57d60477c0452575c5b1983bce3fffac2/openmaptiles/imposm.py#L32-L33 ...
https://github.com/openmaptiles/openmaptiles-tools/blob/c310d1a57d60477c0452575c5b1983bce3fffac2/openmaptiles/imposm.py#L56-L66


TomPohys:
Hi @zstadler, thanks for PR. As I understand `slice_language_tags`, now it tries to find all name:{lang} where lang is the name of the language contained in openmaptiles.yaml. 

What is the reason to add all *name* tag into hstore and not only languages or user-defined values? What effect it will have on the size of the final mbtiles?

Thanks

zstadler:
Hi @TomPohys,

I feel some information is needed on this PR beyond what I initially wrote. I hope this will also address your comment.

#### Impact on contents
Adding user-defined tags to the `tags: hstore` is not covered by this PR. It is done _today_ by adding them to the `mapping.yaml` file of the appropriate layer. For example, tags I've added to `layers/transportation/mapping.yaml`:
```yaml
tags:
  include:
    - mtb:name
    - mtb:name:en
    - mtb:name:he
    - mtb:oneway
    - maxspeed
```

While I wanted the `mtb:name*` tags to show in the `transportation_name` layer, I found that `slice_language_tags` limits the tags that are passed to the layer:

https://github.com/openmaptiles/openmaptiles-tools/blob/223dd075827f74bac4b3b33fc1da89d9a0e4860f/openmaptiles/sql.py#L88-L93

`slice_language_tags` determines the set of tags used by the "merge by name" mechanism of the `transportation_name` and `waterway` layers, as recently updated in https://github.com/openmaptiles/openmaptiles/pull/989.

With this PR, all 3 `mtb:name*` values in the example above will be included in the `transportation_name` layer, and considered by the "merge by name" mechanism.

#### Impact on size
I do not expect this PR to have any effect on the size of the resulting mbtiles file, especially if no user-defined tags are added. If a user adds `*name*` tags, the size will be minimally increased to include the new tags and their _desired_ impact on the "merge by name" mechanism.

Similarly, I do not expect this PR to have any effect on the size of the PG DB. As far as I understand PG, using the `hstore` column for adding _rare_ tags, such as `mtb:name:en` or `seamark:name`, is better than creating new columns. I suppose this is also the rational for using it for the languages listed in `openmaptiles.yaml`.

TomPohys:
Hi @zstadler, thank you for the detailed explanation and PR.

TomPohys (approved):

zstadler:
@TomPohys,
Could you merge it?

zstadler:
Thanks!

PR #316, opened 2020-09-17
WIP: generate-vectortiles: node 12 supports, drop filter_deprecation

smellman:
In my case, only AMD Ryzen machine will fail in progress generate-tiles.
So I try to bump node 5 to node 12.

- bump all versions.
- mapnik to 4.5.2 from 3.x in all dependencies.
- dropped filter_deprecation because the function break with new tilelive copy.

But there are some problems...

- tl, tilelive-tmsource are no test code.
- tilelive-mapnik's test will fail so I need more research about it.

[tilelive ecosystems](https://github.com/mapbox/TileLive) is not active now.
so I recommend to fork into [OMT](https://github.com/openmaptiles) and maintain forks.

### DIFF
--- a/docker/generate-vectortiles/Dockerfile
+++ b/docker/generate-vectortiles/Dockerfile
@@ -1,16 +1,25 @@
-FROM node:5
+FROM node:12
 MAINTAINER Lukas Martinelli <me@lukasmartinelli.ch>
+
+# to resolve sqlite3 permission errors
+USER node
+RUN mkdir -p /home/node/.npm-global ; \
+    chown -R node:node /home/node/.npm-global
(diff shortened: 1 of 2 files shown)


nyurik:
do we really want to support Mapnik-based tile generation going forward?  Seems too much of a problem.

smellman:
@nyurik I think too and I'm looking for alternative.
Also, my friend created produce planet program, how about it?
https://github.com/un-vector-tile-toolkit/produce-320/

smellman:
My server fail in progress generate-tiles again.

smellman:
I close this because #317 seems good than my approach.

PR #317, opened 2020-09-17
Pr pgquery

lazaa32:
generate-tile script moved to /bin folder. Updated main Dockerfile.


TomPohys (approved):

nyurik (changes requested):

nyurik:
@lazaa32 looks like it is almost ready!  I added a few minor comments, should be easy to fix, and ready to merge :)

bin/generate-tiles
```diff
@@ -0,0 +1,50 @@
+#!/bin/bash
+set -o errexit
+set -o pipefail
+set -o nounset
+
+# For backward compatibility, allow both PG* and POSTGRES_* forms,
+# with the non-standard POSTGRES_* form taking precedence.
+# An error will be raised if neither form is given, except for the PGPORT
+export PGHOST="${POSTGRES_HOST:-${PGHOST?}}" # "xxx.xxx.xxx.xxx&host=xxx.xxx.xxx.xxx&host=..."
```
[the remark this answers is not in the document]
nyurik: What I meant above is that we may want to give it a different name, e.g. PGHOSTS or something like that?  This way all other tools will continue to work properly with the original single-host value of PGHOST, while this tool will be able to use all of them.  The possible logic could be:

* if `PGHOSTS` is set, use it and compute `HOST_COUNT` by counting the `&` + 1
* if `PGHOSTS` is not set, use `PGHOST` and set `HOST_COUNT` to 1

bin/generate-tiles
```diff
@@ -0,0 +1,50 @@
...
+export PGDATABASE="${POSTGRES_DB:-${PGDATABASE?}}"
+export PGUSER="${POSTGRES_USER:-${PGUSER?}}"
+export PGPASSWORD="${POSTGRES_PASSWORD:-${PGPASSWORD?}}"
+export PGPORT="${POSTGRES_PORT:-${PGPORT:-5432}}"
+
+export FUNCZXY="${FUNCZXY:-getmvt}"
+export CPU=${CPU:-4}  # number of CPUs per postgres server
```
smellman, line 16: I suggest `COPY_CONCURRENCY` instead of `CPU`.
https://github.com/openmaptiles/openmaptiles/blob/master/.env#L49

bin/generate-tiles
```diff
@@ -0,0 +1,50 @@
...
+
+export EXPORT_DIR=${EXPORT_DIR:-/export}
+export MBTILES_FILE=${MBTILES_FILE:-tiles.mbtiles}
+export RENDER_SCHEME=${RENDER_SCHEME:-pyramid}
+export RETRY=${RETRY:-2}
+export BBOX=${BBOX:-"-180.0,-85.0511,180.0,85.0511"}
+export TIMEOUT=${TIMEOUT:-1800000}
+export MIN_ZOOM=${MIN_ZOOM:-0}
+export MAX_ZOOM=${MAX_ZOOM:-14}
+export HOST_COUNT=${HOST_COUNT:-1}  # number of postgres servers
```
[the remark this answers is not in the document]
lazaa32: `HOST_COUNT` calculated from `PGHOST` or `PGHOSTS_LIST`.

bin/generate-tiles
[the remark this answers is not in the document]
lazaa32: I introduced `PGHOSTS_LIST` so it is more obvious there are multiple hosts. Logic to set `HOST_COUNT` set as you had suggested.

bin/generate-tiles
```diff
@@ -0,0 +1,50 @@
...
+export ALL_STREAMS=$(( MAXCONNECTIONS * HOST_COUNT ))
+
+export PGQUERY="pgquery://?database=$PGDATABASE&host=$PGHOST&port=$PGPORT&username=$PGUSER&password=$PGPASSWORD&funcZXY=$FUNCZXY&testOnStartup=false&maxpool=$MAXCONNECTIONS&nogzip=$NO_GZIP&key=$KEY"
```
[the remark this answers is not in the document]
lazaa32: Braces added.

bin/generate-tiles
lazaa32: I used `COPY_CONCURRENCY` instead of `CPU`. I don't know what should I set for default value though. Set `1` for now.

lazaa32:
@nyurik @smellman I added your requested changes. Please review it. Thanks!

bin/generate-tiles
```diff
@@ -0,0 +1,69 @@
+#!/bin/bash
+set -o errexit
+set -o pipefail
+set -o nounset
+
+# For backward compatibility, allow both PG* and POSTGRES_* forms,
+# with the non-standard POSTGRES_* form taking precedence.
+# An error will be raised if neither form is given, except for the PGPORT
+export PGHOST="${POSTGRES_HOST:-${PGHOST?}}"
```
nyurik, line 9: the way it is written now, you require that both `POSTGRES_HOST/PGHOST` and `PGHOST_LIST` be set if the user wants to use multiple hosts.  I think you should move the above line to the `else` block below -- i.e. when `PGHOST_LIST` is not set.

bin/generate-tiles
```diff
@@ -0,0 +1,69 @@
...
+else
+  export HOST_COUNT=`awk -F"&" '{print NF}' <<< "${PGHOSTS_LIST}"`
+  export PGHOSTS="${PGHOSTS_LIST}"
+fi
+
+export NO_GZIP=${NO_GZIP:-1}
+if [[ ${NO_GZIP} ]] && [[ ! ${NO_GZIP} == "0" ]] && [[ ! ${NO_GZIP} == "true" ]]
+then
+  export NO_GZIP=1
+fi
```
nyurik, line 29: AFAIK, you should never use vars like this without quotes.  Something like this perhaps?  And same for the KEY (btw, we may want to rename KEY -- too confusing -- how about `USE_KEY` or `USE_KEY_COLUMN` ?

```suggestion
if [[ -n "${NO_GZIP:}" ]] && [[ "${NO_GZIP:}" != "0" ]] && [[ "${NO_GZIP:}" != "false" ]]; then
  export NO_GZIP=1
else
  export NO_GZIP=
fi
```

bin/generate-tiles
```diff
@@ -0,0 +1,69 @@
...
+
+export KEY=${KEY:-1}
+if [[ ${KEY} ]] && [[ ! ${KEY} == "0" ]] && [[ ! ${KEY} == "true" ]]
+then
+  export KEY=1
+fi
+
+export FUNCZXY="${FUNCZXY:-getmvt}"
```
nyurik, line 37:
```suggestion
export FUNC_ZXY="${FUNC_ZXY:-getmvt}"
```

bin/generate-tiles
```diff
@@ -0,0 +1,69 @@
...
+export EXPORT_DIR=${EXPORT_DIR:-/export}
+export MBTILES_FILE=${MBTILES_FILE:-tiles.mbtiles}
+export RENDER_SCHEME=${RENDER_SCHEME:-pyramid}
+export RETRY=${RETRY:-2}
+export BBOX=${BBOX:-"-180.0,-85.0511,180.0,85.0511"}
+export TIMEOUT=${TIMEOUT:-1800000}
+export MIN_ZOOM=${MIN_ZOOM:-0}
+export MAX_ZOOM=${MAX_ZOOM:-14}
+
+export PGQUERY="pgquery://?database=${PGDATABASE}&host=${PGHOST}&port=${PGPORT}&username=${PGUSER}&password=${PGPASSWORD}&funcZXY=${FUNCZXY}&testOnStartup=false&maxpool=${MAXCONNECTIONS}&nogzip=${NO_GZIP}&key=${KEY}"
```
nyurik, line 51: testOnStartup is a highly useful thing to have in many cases -- lets make that into a var too, with default = true

nyurik (changes requested):
thanks, a few more comments and a bug

bin/generate-tiles
lazaa32: Oh, yes. You are right. I didn't realize the user might have set just `PGHOSTS_LIST`

bin/generate-tiles
lazaa32: +1 for `USE_KEY_COLUMN`. It seems to me more descriptive.

bin/generate-tiles
lazaa32: `TEST_ON_STARTUP` var added.

bin/generate-tiles
```diff
@@ -0,0 +1,69 @@
...
+# An error will be raised if neither form is given, except for the PGPORT
+export PGHOST="${POSTGRES_HOST:-${PGHOST?}}"
+export PGDATABASE="${POSTGRES_DB:-${PGDATABASE?}}"
+export PGUSER="${POSTGRES_USER:-${PGUSER?}}"
+export PGPASSWORD="${POSTGRES_PASSWORD:-${PGPASSWORD?}}"
+export PGPORT="${POSTGRES_PORT:-${PGPORT:-5432}}"
+
+# List of postgres servers
+if [[ -z "${PGHOSTS_LIST}" ]] # "xxx.xxx.xxx.xxx&host=xxx.xxx.xxx.xxx&host=..."
+then
```
nyurik, line 17: ah, and one more thing - just for consistency with all other scripts, could you put those on a single line?  
```suggestion
if [[ -z "${PGHOSTS_LIST}" ]] # "xxx.xxx.xxx.xxx&host=xxx.xxx.xxx.xxx&host=..."; then
```

bin/generate-tiles
lazaa32: @nyurik just not sure what is the purpose of colon in `"${NO_GZIP:}`. When I test it I get `bash: ${NO_GZIP:}: bad substitution`

bin/generate-tiles
```diff
@@ -0,0 +1,69 @@
...
+
+export KEY=${KEY:-1}
+if [[ ${KEY} ]] && [[ ! ${KEY} == "0" ]] && [[ ! ${KEY} == "true" ]]
+then
+  export KEY=1
+fi
+
+export FUNCZXY="${FUNCZXY:-getmvt}"
+export COPY_CONCURRENCY=${COPY_CONCURRENCY:-1}  # number of CPUs per postgres server
+export MAXCONNECTIONS=${MAXCONNECTIONS:-${COPY_CONCURRENCY}}
```
nyurik, line 39: We should clarify MAX_HOST_CONNECTIONS is per host.  Sorry all the nitpicking, but naming is hard -- once we merge, we won't be able to easily change those :)

```suggestion
export MAX_HOST_CONNECTIONS=${MAX_HOST_CONNECTIONS:-${COPY_CONCURRENCY}}
```

bin/generate-tiles
```diff
@@ -0,0 +1,69 @@
...
+        --timeout="$TIMEOUT" \
+        --minzoom="$MIN_ZOOM" \
+        --maxzoom="$MAX_ZOOM" \
+        --concurrency="$ALL_STREAMS" \
+        "$PGQUERY" \
+        "mbtiles://${EXPORT_DIR}/${MBTILES_FILE}"
+}
+
+#echo "Generating zoom from $MIN_ZOOM to $MAX_ZOOM."
+echo "Generating zoom $MIN_ZOOM..$MAX_ZOOM from $HOST_COUNT servers, using $MAXCONNECTIONS connections per server, $ALL_STREAMS streams"
```
nyurik, line 68:
```suggestion
echo "Generating zoom $MIN_ZOOM..$MAX_ZOOM from $HOST_COUNT servers, using $MAXCONNECTIONS connections per server, $ALL_STREAMS parallel streams"
```

bin/generate-tiles
lazaa32: Formatting done.

bin/generate-tiles
lazaa32: Added to echo.

bin/generate-tiles
lazaa32: Renamed.

bin/generate-tiles
```diff
@@ -0,0 +1,50 @@
...
+export PGHOST="${POSTGRES_HOST:-${PGHOST?}}" # "xxx.xxx.xxx.xxx&host=xxx.xxx.xxx.xxx&host=..."
+export PGDATABASE="${POSTGRES_DB:-${PGDATABASE?}}"
+export PGUSER="${POSTGRES_USER:-${PGUSER?}}"
+export PGPASSWORD="${POSTGRES_PASSWORD:-${PGPASSWORD?}}"
+export PGPORT="${POSTGRES_PORT:-${PGPORT:-5432}}"
+
+export FUNCZXY="${FUNCZXY:-getmvt}"
+export CPU=${CPU:-4}  # number of CPUs per postgres server
+export MAXCONNECTIONS=${MAXCONNECTIONS:-${CPU}}
+export NO_GZIP=${NO_GZIP:-1}
```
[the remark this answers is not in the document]
lazaa32: Set default to 1. Logic added.

bin/generate-tiles
lazaa32: It also seems to me that no matter what you have set for NO_GZIP it always get reset to either empty or 1. Even though you have it set for true or 0.

bin/generate-tiles
```diff
@@ -0,0 +1,70 @@
...
+        --minzoom="$MIN_ZOOM" \
+        --maxzoom="$MAX_ZOOM" \
+        --concurrency="$ALL_STREAMS" \
+        "$PGQUERY" \
+        "mbtiles://${EXPORT_DIR}/${MBTILES_FILE}"
+}
+
+#echo "Generating zoom from $MIN_ZOOM to $MAX_ZOOM."
+echo "Generating zoom $MIN_ZOOM..$MAX_ZOOM from $HOST_COUNT servers, using $MAX_HOST_CONNECTIONS connections per server, $ALL_STREAMS parallel streams..."
+#export_local_mbtiles
```
zstadler, line 70: Is this line commented-out by mistake?
```suggestion
export_local_mbtiles
```

lazaa32: Thanks, it was a mistake.

bin/generate-tiles
lazaa32: I had a similar problem as nyurik/tilelive-pgquery#1 Maybe set default to false?

bin/generate-tiles
```diff
@@ -0,0 +1,69 @@
...
+fi
+
+export USE_KEY_COLUMN=${USE_KEY_COLUMN:-1}
+if [[ ${USE_KEY_COLUMN} ]] && [[ ! ${USE_KEY_COLUMN} == "0" ]] && [[ ! ${USE_KEY_COLUMN} == "true" ]]
+then
+  export USE_KEY_COLUMN=1
+fi
+
+export FUNC_ZXY=${FUNC_ZXY:-getmvt}
+export TEST_ON_STARTUP="${TEST_ON_STARTUP:-0}"
```
nyurik, line 38: Should the default be some tile here?  You could use the same one i use for default, e.g. `14/9268/3575`.  It's good to do a test on startup to catch any connection errors.  Note that an empty value is the same as `false`, whereas not setting it would enable it by default.

nyurik (commented):
the test on startup has been fixed, made a relevant suggestion inline

Dockerfile
```diff
@@ -120,6 +120,17 @@ RUN set -eux ;\
...
+    curl -sL https://deb.nodesource.com/setup_12.x | bash -  ;\
+    DEBIAN_FRONTEND=noninteractive apt-get update  ;\
+    DEBIAN_FRONTEND=noninteractive apt-get install -y --no-install-recommends  \
+        nodejs  ;\
+    rm -rf /var/lib/apt/lists/  ;\
+    npm config set unsafe-perm true  ;\
+    npm install -g \
+      @mapbox/mbtiles@0.12.1 \
+      @mapbox/tilelive@6.1.0 \
+      tilelive-pgquery@0.7.2 ;\
```
nyurik, line 133:
```suggestion
      tilelive-pgquery@0.7.3 ;\
```

bin/generate-tiles
nyurik: i fixed it - the issue was that if the value was not set, it worked correctly. But if you do set it, it has to be a proper `z/x/y` - and i had a bug in parsing it. Fixed in 0.7.3

bin/generate-tiles
zstadler: pgquery's default  - `14/9268/3575` - didn't work for me because it is not covered by my download area.
Could `0/0/0` be an option here?

nyurik: `0/0/0` is not a good candidate because it could take over an hour to generate.  It is ok if the downloaded area is not including that tile -- as long as query runs ok, even if it returns an empty result, it shows that everything is working. Plus you can always configure it to test one of the tiles inside your boundaries.

lazaa32: I had the same problem that the test tile wasn't inside my downloaded area. But as long it is not an issue and the query runs and returns empty result that serves only for testing I think we are good. I put `14/9268/3575` as default for now.

Dockerfile
lazaa32: tilelive-pgqeury version updated.

lazaa32:
@nyurik @zstadler I upgraded the version of tilelive-pgquery. Could you please have a look at this one https://github.com/openmaptiles/openmaptiles-tools/pull/317#discussion_r491498299 so we can resolve it? 
 Thanks

bin/generate-tiles
nyurik: @lazaa32 sorry, forgot a dash (fixed in the comment).  The `"${FOO:-}"` would use FOO unless its empty or null, in which case it will use whatever is after the `:-`. In this case, the value is an empty string.  Note that just using `${FOO}` would not work because of the `set -u` above, so you have to be explicit here.

lazaa32: @nyurik thanks, I'm not a really expert in bash so I wondered what that syntax means :D . However it still seems to me that if  you set `NO_GZIP=0` after this condition you end up with `empty`. If you set `NO_GZIP=true` you end up with `NO_GZIP=1`. Am I wrong when i think that boolean works like this `false=1=empty` and `true=0`?

zstadler: The bash manual says:
> **${_parameter_:-_word_}**
     If _parameter_ is _unset_ or _null_, the expansion of _word_ is substituted. Otherwise, the value of _parameter_ is substituted.

Therefore "${FOO}` and `${FOO:-}` are identical

```suggestion
case "${NO_GZIP:-0}" in
  0 | false ) export NO_GZIP= ;;
  * ) export NO_GZIP=1 ;;
esac
```

nyurik: @zstadler this is not correct because `set -u` (same as `set -o nounset`) will trigger an error if you try to reference an undeclared variable.

bin/generate-tiles
```diff
@@ -0,0 +1,69 @@
...
+if [[ ${NO_GZIP} ]] && [[ ! ${NO_GZIP} == "0" ]] && [[ ! ${NO_GZIP} == "true" ]]
+then
+  export NO_GZIP=1
+fi
+
+export USE_KEY_COLUMN=${USE_KEY_COLUMN:-1}
+if [[ ${USE_KEY_COLUMN} ]] && [[ ! ${USE_KEY_COLUMN} == "0" ]] && [[ ! ${USE_KEY_COLUMN} == "true" ]]
+then
+  export USE_KEY_COLUMN=1
+fi
```
zstadler, line 35: Assuming the default for `USE_KEY_COLUMN` should be `USE_KEY_COLUMN=1`

```suggestion
case "${USE_KEY_COLUMN:-1}" in
  0 | false ) export USE_KEY_COLUMN= ;;
  * ) export USE_KEY_COLUMN=1 ;;
esac
```

If the default should be  `USE_KEY_COLUMN=`, use `case "${USE_KEY_COLUMN:-0}" in`

bin/generate-tiles
nyurik: @lazaa32 I think the "cleanest solution" (whatever that even means)  would be to actually set `NO_GZIP` and other boolean vars to the final result. Something like this:

```if ...; then
  export NO_GZIP="&nogzip=1"
else
  export NO_GZIP=
fi

...

PGQUERY="...${NO_GZIP}..."
```

This way we create the exact part of the query as desired. This also means that there is no need to define the value for TEST_ON_STARTUP -- because you won't even specify that value if the user hasn't set it.  What do you think?

zstadler: @nyurik, you are right about `${FOO:-}`

I believe my suggestion makes this logic easier to read.

nyurik: TBH, i'm not in favor of the `case` statement -- much less readable than an `if` with a list of conditions. I'm not even sure if we use case anywhere else. Also, I think it would be better to just declare the whole `&nogzip=1` vs an empty value -- this way you control exactly what you expand.

lazaa32: +1 to keep `if` since we don't use `case` in other scripts. Also exported the final string directly into var.

lazaa32:
@nyurik @zstadler can you please confirm I set the default values right? I like the idea of exporting part of the pgquery string directly into variable, it seems easier to read to me.Any additional comments welcomed.

nyurik (approved):
awesome work, thank you!!!

PR #319, opened 2020-09-23
Allow layer wihtout attribute in yaml mapping

frodrigo:
Current tool fails when there is no attribute field in a layer. It does not support empty `fields:` or no `fields:` at all in the mapping.


TomPohys (approved):

TomPohys:
Thanks @frodrigo.

Issue #300, opened 2020-06-25
import-sql does not stop on error


frodrigo:
Since parallel import is available, when SQL import fails it does not stop.

1. It make hard to find where the error is.
2. Even when `MAX_PARALLEL_PSQL=1`, the import continue after the fail.

PR #322, opened 2020-09-23
Stop parallel import-sql on error #300

frodrigo:
When import all SQL in parallel it does not stop on error.

Use xargs special 255 exit code to stop running new jobs. But the running jobs still continues to the end.

### DIFF
--- a/bin/import-sql
+++ b/bin/import-sql
@@ -47,7 +47,7 @@ function import_all_sql_files() {
       # Run import-sql script in parallel, up to MAX_PARALLEL_PSQL processes at the same time
       : "${MAX_PARALLEL_PSQL:=5}"
       echo "Importing $(find "$dir/parallel" -name "*.sql" | wc -l) sql files from $dir/parallel/, up to $MAX_PARALLEL_PSQL files at the same time"
-      find "$dir/parallel" -name "*.sql" -print0 | xargs -0 -I{} -P "$MAX_PARALLEL_PSQL" "$0" "{}"
+      find "$dir/parallel" -name "*.sql" -print0 | xargs -0 -I{} -P "$MAX_PARALLEL_PSQL" sh -c "\"$0\" \"{}\" || exit 255"
       echo "Finished importing sql files matching '$dir/parallel/*.sql'"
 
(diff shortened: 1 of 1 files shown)


TomPohys (approved):

TomPohys:
Awesome, thanks @frodrigo.

bin/import-sql
```diff
@@ -47,7 +47,7 @@ function import_all_sql_files() {
       # Run import-sql script in parallel, up to MAX_PARALLEL_PSQL processes at the same time
       : "${MAX_PARALLEL_PSQL:=5}"
       echo "Importing $(find "$dir/parallel" -name "*.sql" | wc -l) sql files from $dir/parallel/, up to $MAX_PARALLEL_PSQL files at the same time"
-      find "$dir/parallel" -name "*.sql" -print0 | xargs -0 -I{} -P "$MAX_PARALLEL_PSQL" "$0" "{}"
+      find "$dir/parallel" -name "*.sql" -print0 | xargs -0 -I{} -P "$MAX_PARALLEL_PSQL" sh -c "\"$0\" \"{}\" || exit 255"
```
nyurik, line 50: is it possible to use single quotes here?  If so, would make it cleaner

TomPohys: `sh -c '\"$0\" \"{}\" || exit 255'` rise me an error `"sh": not found`

nyurik: you wouldn't need to escape the double quotes - `sh -c '"$0" "{}" || exit 255'`

frodrigo: No, in this way $0 inside single quote it not interpreted.

nyurik: good point thx.

frodrigo: > `sh -c '\"$0\" \"{}\" || exit 255'` rise me an error `"sh": not found`

No idea why `sh` is missing on your side. I can replace `sh` by `/usr/bin/env bash`?

TomPohys: It was when I test switch to single quotes. No problem as it is now. Thanks

PR #323, opened 2020-09-24
Use pg_roles view instead of pg_authid table

baipi:
Some cloud providers don't allow to query table `pg_authid` due to security reasons (that's the case on CloudSQL and I think it's the same for AWS RDS).

Running `./bin/refresh-views` on a CloudSQL database results with the following error:
```
asyncpg.exceptions.InsufficientPrivilegeError: permission denied for table pg_authid
```

According to [PostgreSQL documentation](https://www.postgresql.org/docs/current/catalog-pg-authid.html):

> Since this catalog contains passwords, it must not be publicly readable. pg_roles is a publicly readable view on pg_authid that blanks out the password field.

This small change fixes the error.


nyurik (approved):
thx!

PR #324, opened 2020-10-01
download-osm: simplify BBOX integration into OMT

zstadler:
By including a `BBOX=` prefix into the `.bbox` file, its contents can be easily integrated into other OMT components.
For example:
- In shell scripts use `. data/${area}.bbox`
- In make files use `include data/${area}.bbox`


nyurik:
@zstadler initially I actually had that and removed it. The .env makes things more difficult on the consumption, because it allows the file to be modified, e.g. you can add a comment and other variables.  The quickstart/makefile need to parse that value, and I don't want to overcomplicate it there, thus it is easier to keep just the value.

zstadler:
@nyurik,

I'm sorry if I have not made myself clear.

This PR is not about adding the machine-generated `BBOX` value to `.env`.
I fully agree that there should not be any machine editing or modification of the user's `.env` file.

This is about adding a `"BBOX="` prefix to the _contents_ of the new `${AREA_BBOX_FILE}`.
The goal to simplify its use in the Makefile, as I suggested at
[https://github.com/openmaptiles/openmaptiles/pull/1012/files#diff-b67911656ef5d18c4ae36cb6741b7965R134](https://github.com/openmaptiles/openmaptiles/pull/1012/files#diff-b67911656ef5d18c4ae36cb6741b7965R134):

```diff
  ifneq (,$(wildcard $(AREA_BBOX_FILE)))
-   cat := $(if $(filter $(OS),Windows_NT),type,cat)
-   BBOX := $(shell $(cat) ${AREA_BBOX_FILE})
+   include $(AREA_BBOX_FILE)
    export BBOX
  endif
```

If anyone wants to manually modify `${AREA_BBOX_FILE}`, they can do that regardless of the `BBOX=` prefix it may contain.

nyurik:
@zstadler apologies for not explaining fully -- I don't mean machine editing of the bbox file, i meant that users could modify it to include other values or comments, and I would rather restrict the meaning of the file to just a single value that is easy to use from everywhere - i.e. you can use it from the makefile, or you can load that value elsewhere (like in some python script) without any extra parsing. Adding `BBOX=` means that in the future someone could add other values to it, and I would rather prevent that.

zstadler:
@nyurik,

Why would a user modify a `${AREA_BBOX_FILE}` in the `data` directory to set additional environment variables, rather than use the well-known `.env` file, which is already included by the `Makefile`?
On the other hand, user modifications of other machine-generated files in the `data` and `build` directories can break the flow as well. What is the motivation to increase the security level of this particular file compared to all others?

PR #326, opened 2020-10-02
Implement osmfr poly bbox support

nyurik:
When downloading from osmfr,
will use .poly files to compute BBOX


zstadler:
Thank you!

TomPohys (approved):

PR #335, opened 2020-11-03
Add mountain_peak_layer for test purpose.

lazaa32:
Related to #333.
I added a `POINT` layer `mountain_peak`. However I am not able to run `make test` nor `make rebuild-expected`. I got always the same error:
```
postgres_1  | 2020-11-03 14:39:29.264 UTC [152] ERROR:  function layer_mountain_peak(geometry, integer, numeric) does not exist at character 367
postgres_1  | 2020-11-03 14:39:29.264 UTC [152] HINT:  No function matches the given name and argument types. You might need to add explicit type casts.
```

Is there any other step I have to do before?

### DIFF
--- a/tests/python/test_import_wikidata.py
+++ b/tests/python/test_import_wikidata.py
@@ -23,7 +23,7 @@ class UtilsTestCase(IsolatedAsyncioTestCase):
 
     async def test_find_tables(self):
         tables = importer.find_tables(test_dir / '../testlayers/testmaptiles.yaml')
-        self.assertEqual(tables, ['osm_housenumber_point'])
+        self.assertEqual(tables, ['osm_housenumber_point', 'osm_peak_point'])
 
     # async def test_pg_func(self):
     #     conn = None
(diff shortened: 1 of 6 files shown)


nyurik:
@lazaa32 i tried to push an update to your branch, but it seems you didn't click the checkbox "allow maintaners to change this pr" (or something similar). Could you merge the latest master?  It would also trigger the recheck.

nyurik:
@lazaa32 as for your original question -- i suspect the reason it is not working is because of this line:

https://github.com/openmaptiles/openmaptiles-tools/pull/335/files#diff-eb20f1dd2e52644a8d296e45ccad76fa4ce7845600d19b530e05489b6baf4061R25

you call `layer_mountain_peak(...)` function, but test sql files are never actually executed on the test database - so that function would not exist there. Note how the existing two layers do not call any non-default functions:

https://github.com/openmaptiles/openmaptiles-tools/blob/e02da05eea22e9f3785b21f8fed542422ee57bf0/tests/testlayers/housenumber/housenumber.yaml#L13

https://github.com/openmaptiles/openmaptiles-tools/blob/e02da05eea22e9f3785b21f8fed542422ee57bf0/tests/testlayers/enumfield/enumfield.yaml#L57

I would suggest you simply fake it in the same manner. That said, what specific tools functionality are you trying to test here?

lazaa32:
Closing. Subsequent discussion in PR #343

PR #340, opened 2021-03-22
Bump PostGIS and GEOS

TomPohys:
Bump PostGIS from 3.0 to 3.1 and GEOS from 3.7.1 to 3.9.0

FIX https://github.com/openmaptiles/openmaptiles/issues/1058

Thanks @suur (https://github.com/openmaptiles/openmaptiles/issues/1058#issuecomment-751360970)


docker/postgis/Dockerfile
```diff
@@ -95,4 +106,4 @@ RUN set -eux  ;\
 
 # The script should run after the parent's 10_postgis.sh runs
 # so it must have the name that's listed after that.
-COPY ./initdb-postgis.sh /docker-entrypoint-initdb.d/20_omt_postgis.sh
+COPY ./initdb-postgis.sh /docker-entrypoint-initdb.d/20_omt_postgis.sh
```
nyurik, line 109: any reason why you are deleting trailing newline?  I think your editor needs to be configured to always add it -- makes future diffing much easier.

nyurik (changes requested):
I am not sure this is the right approach.  Postgis docker image sets up GEOS, so it would be far better to let postgis docker team to manage this, and possibly submit a pull request to them to update it.  See https://github.com/postgis/docker-postgis/blob/master/12-master/Dockerfile#L103

docker/postgis/Dockerfile
TomPohys: Thanks, I add it.

nyurik:
@TomPohys any thoughts about my prev comment?

TomPohys:
Yes, It would be a lot nicer to have it in the original PostGIS Docker image. Thanks to point it out.

nyurik (approved):
Looks good, thx!

TomPohys:
It looks like, that PostGIS 9.6-3.1 still using GEOS 3.7.1
I don't understand why, if the commit hash is pointing to https://github.com/libgeos/geos/commit/32b99ca05b99b733dcd2c274c5f14e6402e6e5a1.

`docker run --name some-postgis -e POSTGRES_PASSWORD= -d postgis/postgis:9.6-3.1`
`docker run -it --link some-postgis:postgres --rm postgres \
     sh -c 'exec psql -h "$POSTGRES_PORT_5432_TCP_ADDR" -p "$POSTGRES_PORT_5432_TCP_PORT" -U postgres'`
```
psql (13.2 (Debian 13.2-1.pgdg100+1), server 9.6.21)
Type "help" for help.

postgres=# select postgis_full_version();
                                                                                           postgis_full_version                                                                                           
----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
 POSTGIS="3.1.1 aaf4c79" [EXTENSION] PGSQL="96" GEOS="3.7.1-CAPI-1.11.1 27a5e771" PROJ="Rel. 4.9.3, 15 August 2016" LIBXML="2.9.4" LIBJSON="0.12.1" LIBPROTOBUF="1.2.1" WAGYU="0.5.0 (Internal)" TOPOLOGY
(1 row)
```

nyurik:
@TomPohys is there an issue or a pull request in the https://github.com/postgis/docker-postgis repo?

TomPohys:
I thought that this commit https://github.com/postgis/docker-postgis/commit/c58090124ccd7f54b39bb5bcf2f79b7011e94964 to update GEOS to a newer version. And if in dockerfile of `13-master` is 
```
# geos
ENV GEOS_VERSION master
ENV GEOS_GIT_HASH 32b99ca05b99b733dcd2c274c5f14e6402e6e5a1
```
https://github.com/postgis/docker-postgis/blob/master/13-master/Dockerfile#L98

I thought that that was a proper change to bump up the GEOS

nyurik:
I think that only applies to `12-master` and `13-master` tags (i might be wrong)

nyurik:
I recommend starting an issue in the docker-postgis repo -- to make sure it is properly brought to the attention of the maintainers.

PR #342, opened 2021-03-30
download-osm better error handling

nyurik:
* print an error if pbf file does not exist for bbox command
* print more debug info on verbose


TomPohys (approved):

TomPohys:
Thanks!

PR #345, opened 2021-04-15
Bump GEOS 3.9.1

TomPohys:
Bump GEOS 3.9.1

```
postgres=# select postgis_full_version();
                                                                                  postgis_full_version                                                                                  
----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
 POSTGIS="3.1.1 aaf4c79" [EXTENSION] PGSQL="96" GEOS="3.9.1-CAPI-1.14.2" PROJ="Rel. 4.9.3, 15 August 2016" LIBXML="2.9.4" LIBJSON="0.12.1" LIBPROTOBUF="1.2.1" WAGYU="0.5.0 (Internal)"
```


docker/postgis/Dockerfile
```diff
@@ -36,6 +36,16 @@ RUN set -eux  ;\
...
+    ## GEOS
+    cd /opt/ ;\
+    curl -o /opt/geos.tar.bz2 http://download.osgeo.org/geos/geos-3.9.1.tar.bz2 ;\
+    mkdir /opt/geos ;\
+    tar xf /opt/geos.tar.bz2 -C /opt/geos --strip-components=1 ;\
+    cd /opt/geos/ ;\
+    ./configure ;\
+    make -j ;\
+    make install ;\
+    rm -rf /opt/geos* ;\
```
nyurik, line 48: please add this line before line 10:

```
ARG GEOS_VER=3.9.1
```

and add an extra space for consistency with other code (plus use the above var):

```suggestion
    ## GEOS
    cd /opt/  ;\
    curl -o /opt/geos.tar.bz2 http://download.osgeo.org/geos/geos-${GEOS_VER}.tar.bz2  ;\
    mkdir /opt/geos  ;\
    tar xf /opt/geos.tar.bz2 -C /opt/geos --strip-components=1  ;\
    cd /opt/geos/  ;\
    ./configure  ;\
    make -j  ;\
    make install  ;\
    rm -rf /opt/geos*  ;\
```

TomPohys: Done. Thanks.

TomPohys:
Some function of GEOS 3.9.1 will not be applicable (https://github.com/postgis/docker-postgis/issues/233#issuecomment-821763441)

nyurik:
Sigh, thus is really beginning to feel like a hack - postgis is compiled against the old version,  but then we swap geoslib to a new version underneath it, hoping that the lib is perfectly backwards compatible, which might not be the case... at the same time,  manual compilation of postgis itself is even worse...

So let's continue with this patch for testing, but only keep it in master, or maybe release it as a side branch, but let's try to work with @ImreSamu on the upstream postgis docker image, and use that for releases

TomPohys:
Yes, I agree to keep it in the `master` branch for just the `latest` docker image. After release Debian Bullseye with GEOS 3.9.0 we can remove the force GEOS update.

lazaa32 (approved):

PR #355, opened 2021-08-05
Add "copy" command to mbtiles-tools

nyurik:
Copy tiles from one mbtiles file to another. Copying can be limited by one or more zooms, and a bounding box. This action also runs meta-copy command.

Extras:
 * cleaned up zoom parameter parsing in several tools
 * added saving of the build dir as a build artifact when running github action on pull request

cc @nickpeihl and also Martin Guillon was asking about it a while back - can't find his github account though


openmaptiles/mbtile_tools.py
```diff
@@ -261,8 +262,8 @@ async def generate(self, tileset, reset, auto_minmax,
             f'Connecting to PostgreSQL at {pghost}:{pgport}, db={dbname}, user={user}...')
         try:
             async with asyncpg.create_pool(
-                database=dbname, host=pghost, port=pgport, user=user,
-                password=password, min_size=1, max_size=1,
+                    database=dbname, host=pghost, port=pgport, user=user,
+                    password=password, min_size=1, max_size=1,
```
zstadler, line 266: Is this additional indentation intentional?

nyurik: Yes, parameters to a function should be indented more than a block.  When i did a file reformat, these got changed.

nickpeihl (approved):
LGTM, thanks!

`sqldiff` reports some differences, but `geocoder_data`, `grid_key`, and `keymap` tables are empty in my source database and not required [per the mbtiles spec](https://github.com/mapbox/mbtiles-spec/blob/master/1.3/spec.md). I don't think those tables are commonly used either. So probably fine to ignore these tables unless someone needs them.

```
$ sqldiff <target.mbtiles> <source.mbtiles> --schema
CREATE TABLE geocoder_data (
    type TEXT,
    shard INTEGER,
    data BLOB
);
CREATE INDEX geocoder_type_index ON geocoder_data (type);
CREATE UNIQUE INDEX geocoder_shard_index ON geocoder_data (type, shard);
CREATE TABLE grid_key (
    grid_id TEXT,
    key_name TEXT
);
CREATE UNIQUE INDEX grid_key_lookup ON grid_key (grid_id, key_name);
CREATE TABLE grid_utfgrid (
    grid_id TEXT,
    grid_utfgrid BLOB
);
CREATE UNIQUE INDEX grid_utfgrid_lookup ON grid_utfgrid (grid_id);
CREATE TABLE keymap (
    key_name TEXT,
    key_json TEXT
);
CREATE UNIQUE INDEX keymap_lookup ON keymap (key_name);
ALTER TABLE map ADD COLUMN grid_id;
CREATE INDEX map_grid_id ON map (grid_id);
```

Out of curiosity, I also tested this against the Adria raster tile database sample from https://wiki.openstreetmap.org/wiki/MBTiles which did not work as the raster tile database does not have `map` and `images` tables. But I think `mbtiles-tools` is only intended to work with vector tile databases, right? So probably not worth considering raster tile support unless someone specifically needs it.

nickpeihl:
Thanks for clarification @nyurik. Reiterating here that this PR looks good to me and fulfills my need.

nyurik:
I think i will submit another PR for this change, automatically setting the sqlite block size to 512 (same as what the mbtiles planet file is set to). There is a bigger change possible - to allow for an optimized storage mode, but this would warrant a bigger discussion.

nyurik:
I just implemented bbox support as well, so now users can copy a subset of tiles related to just one area. The system handles the whole pyramid filtering, not just a single zoom.  @TomPohys please review and i will merge right after. Thx!

nyurik:
P.S. I made a change to the build action so that it saves all build artifacts -- this way it makes it easier to debug failed builds.

TomPohys (approved):

TomPohys:
Awesome!

Falke-Design:
@nyurik is it possible to merge mbtiles with this? If yes it would be nice to make this more clear in the docu / Readme. 

I'm not the only one who didn't get that with `mbtiles-tools copy` mbtiles can be merged: https://github.com/openmaptiles/openmaptiles/issues/290#issuecomment-932197352

PR #362, opened 2021-08-28
Clean any additional unit information

marq24:
some of the osm fields like the height unfortunately contain units (they should not be there - right)... by adding \w to the last part of the RegEx we can get rid of any kind of additional unit information (no matter if it's m, meter, ft, miles or any combination) - this will result in better 3D building rendering.

An example can be found here https://www.openstreetmap.org/edit#map=19/52.01690/8.52986 (roof of the church) that currently will not be rendered with the appropriate height - you can check it here: https://www.maptiler.com/maps/#streets/vector/18.2/8.529948/52.017556/0.00/60.0

And yes - I understand that "garbage in will result in garbage out" by default ;-) But sometimes this can be handled

here is a shot of my local result
image

### DIFF
--- a/sql/CleanNumeric.sql
+++ b/sql/CleanNumeric.sql
@@ -11,7 +11,7 @@ __Returns:__ `numeric`
 ******************************************************************************/
 create or replace function CleanNumeric (i text) returns numeric as
 $body$
- SELECT substring(i from '^\s*([-+]?(?=\d|\.\d)\d*(?:\.\d*)?(?:[Ee][-+]?\d+)?)\s*$')::numeric;
+ SELECT substring(i from '^\s*([-+]?(?=\d|\.\d)\d*(?:\.\d*)?(?:[Ee][-+]?\d+)?)[\s\w]*$')::numeric;
 $body$
 language sql
 strict immutable cost 20


TomPohys:
Hi @marq24, thank you for this PR. 

Could you please add test for e.g. `17 m` into [tests/sql/CleanNumeric.sql](https://github.com/openmaptiles/openmaptiles-tools/blob/master/tests/sql/CleanNumeric.sql) file and create new `expected` files (steps in https://github.com/openmaptiles/openmaptiles-tools#tools-development)?

Thank you very much

marq24:
I hope this two changes will do it - I have to admit in the 'tests/sql/CleanNumeric.sql' file I haven't understood the logic behind the values behind the '--' ?

marq24:
I see/understand now, WHY my suggested PR might not be THAT smart - since some of your previous 'failed' tests are now valid - I tried to alter the RegEx in order to make sure to expect a 'space' between number and possibly trimmed unit's - but that was not successful.

So at the end yo have to device, if you will accept additional "possible" false positive values when parsing osm number data

nyurik:
@marq tbh, I would rather clean up the values in OSM -- they are editable afterall :) -- rather than try to account for every possible incorrect value. If some value is hugely popular, than we should explicitly handle it, e.g. if people often write `10 ft` we could handle `ft` unit explicitly

marq24:
Cleaning the OSM data manually - This simplified query (over a OSM planet import)
`select * from osm_building_polygon where height SIMILAR TO '\d+? \w*?' and building = 'yes'`
returning 381476 rows - all over the world - so cleaning the source OSM data is IMHO not the right approach even if I fully understand, that "clean" OSM data would be the preferred... I am not sure how many OSM editors we have in the world - and how much time you would like to invest to teach every editor, that units are not permitted in the height field?

When I pick up your train of thought correctly, then attempting to interpret specific units (like ft), implies that you then need to translate the feet value to meter - I know that my current approach have this flaw too - If you know if a value is yards, miles, feets, inchs or centimetres, then I would 'expect' that there is also a translation. Unfortunately the amount of different used units and abbreviation is extremely high (including spelling errors) - so I would not go that route inside sql... (in the openroutingservice project we do something similar in the java code part of the osm data interpretation).

At the end of the day - it's your project - you have to decide how strict you would like to be concerning the input data and what kind of negavive impact you expect from this change - my local version is adjusted and I am very happy with the results.

PR #370, opened 2021-09-11
Add check if required functions and tables for layer are existing

Falke-Design:
Add support for layers to require external SQL functions and tables.
Fixes https://github.com/openmaptiles/openmaptiles/issues/1220

This PR changes the meaning of the `layer / requires` field in the `<layer>.yaml` file:
```
layer:
  requires:
     # each value can be a string or a list of strings
     layers: 'building'
     tables: ['table1','table2']
     functions: 'fnc1(TEXT, TEXT)'
```

For backward compatibility, `requires` can still be a string or a list of strings, in which case that value will be treated as a list of required layers.

Sample SQL that gets generated before all schema sql. PostgreSQL will raise an error if the required table/function is missing.

```
-- Assert table1 exists
SELECT 'table1'::regclass;

-- Assert fnc1 exists
SELECT 'fnc1(TEXT, TEXT)'::regprocedure;
```


openmaptiles/sql.py
```diff
@@ -225,4 +250,19 @@ def field_map(match):
         sql = re.sub(
             r'/\*\s*DELAY_MATERIALIZED_VIEW_CREATION\s*\*/', ' WITH NO DATA ', sql)
 
+    exist_check = ""
```
nyurik, line 253: I think it is better to just add all this at the top of the function as a "result" var, and keep appending to it. Also, we can use a much simpler `SELECT '<table>'::regclass;`  here instead. 

```
     result = ''
     for table in layer.required_tables:
         result += f"-- Assert {table} exists\nSELECT '{table}'::regclass;\n\n"
     for func in layer.required_functions:
         result += f"-- Assert {func} exists\nSELECT '{func}'::regprocedure;\n\n"
     ...
     result += sql
     return result
```

openmaptiles/sql.py
```diff
@@ -117,6 +119,29 @@ def get_slice_language_tags(tileset):
 $$ LANGUAGE SQL IMMUTABLE;
 """
 
+def create_fnc_is_table_existing():
```
nyurik, line 122: we probably should not create global functions that are only needed during the import

nyurik (changes requested):
Thanks for working on this!  There are a few things I think need changing, but this is a great start.

* I don't think we should create global functions just to validate something. All SQL here is dynamically generated, so we can write out any validation logic without using functions. For example, `SELECT 'table'::regclass;` would validate that the table exists and raise an error if it does not.  Also, validating functions should include function signature, e.g. `myfunc(TEXT)` See comment with proposed code change.
* we need to validate yaml during the initial parsing
* naming - i'm not sure what we should name these vars. One idea is to call them   `external_tables` -- implying that these tables come from another layer. We already use  `requires` to indicate external **layers**, so having `required_tables` might be confusing? Or not.

Falke-Design:
How should validating while initial parsing work? At this moment no tables / functions are generated, so it can't checked if they are existing. The checks can only be made while calling the sql of the layer and all steps before are done (like create table).

What if we make a breaking change and make the structure like that:
```
required:
  layers: []
  tables: []
  functions: []
```

nyurik:
By "validate" i meant check that yaml is correct. `Layer.__init__` does all yaml validation. I just realized that all properties are also tested this way using `validate_properties()`.  So lets keep the properties, but maybe just name them a little bit different - there should be no `get` in a property name (properties are like nouns, and functions are like verbs) - `requires_tables`, `requires_functions`.

I'm ok to redefine `requires` to be a dictionary. We can easily make it backwards compatible -- if `requires` is a string, treat it as a list with a single element.  If `requires` is a list, treat it as a dictionary with a list of layers.  See how `__init__` parses `requires`.

nyurik (commented):
Thanks, I added a few more comments.  Overall, I think it will be simpler to create a helper function that tests and normalizes just one property, and call it from the `__init__`:

```
requires = self.definition['layer'].get('requires', {})
if not isinstance(requires, dict):
    requires['layers'] = requires
else:
    requires = requires.copy()  # dict will be modified to detect unrecognized properties

err = "If set, 'requires' parameter must be a map with optional 'layers', 'tables', and 'functions' sub-elements. Each sub-element must be a string or a list of strings. If 'requires' is a list or a string itself, it is treated as a list of layers. "

self.requires_layers = get_requires_prop(requires, 'layers', err + "'requires.layers' must be an ID of another layer, or a list of layer IDs.")
self.requires_tables = get_requires_prop(requires, 'tables', err + "'requires.tables' must be the name of a PostgreSQL table or a view, or a list of tables/views")
self.requires_functions = get_requires_prop(requires, 'functions', err + "'requires.functions' must be a PostgreSQL function name with parameters or a list of functions. Example: 'myfunc(TEXT, TEXT)' ")

if requires:
  # get_requires_prop will delete the key it handled. Remaining keys are errors.
  raise ValueError("Unrecognized sub-elements in the 'requires' parameter:" + +str(list(requires.keys()))
```

Falke-Design:
I will change it.
Do you now want that the layer.yaml looks like that:
```
requires:
  layers: []
  tables: []
  functions: []
```
or that?:
```
requires: deprecated warning and internal mapping to requires_layers
requires_layers: []
requires_tables: []
requires_functions: []
```

What I not really understand is why there are `@property`s if we don't define all possible properties? Like you says `requires_*` should now checked in `__init__` but then they are not defined as property anymore?

nyurik:
sorry for confusion - I think it is cleaner to have YAML as a subtree -- makes it more readable and easier to change:

```
requires:
  layers: []
  tables: []
  functions: []
```

but in python, it is better to have dedicated class fields for each one -- fields are easier to work with and more predictable than having a generic dictionary - hence `requires_layers`, ...

There is little difference between a `@property` and a class field from the usage perspective -- both are used with a dot -- `layer.some_property`. I think in this specific case a field is better than a property because you want to validate that `requires` has only known sub-elements, so might as well do it all at once, but you are welcome to do them as properties too.

Falke-Design:
I updated now the code and the wanted chages. Thanks for the review, it is now much cleaner! 

In my eyes I would change one thing. You suggested to move the requires parsing into the `__init__` I think it would be better to do this in a own function too. Because now we have this code twice, one time for `tileset.py` and the other one in the `test_sql.py`.
With a own function it would be possible to call it from `test_sql.py`.

Also I tried to add tests if it throws Exception with wrong inputs but it was not possible for me.

nyurik:
why do you need to duplicate code in a test?

Falke-Design:
In the test_sql.py the function `query` needs the parsed `requires` fields to be added to the `Case`. Currently the only way to parse requires is to copy the `__init__` logic, because there are tests with the old, new, combined `requires` field.

It would be cleaner to get the parsed data from the tileset-layer and add it to the Case but i have not found a way. (Maybe the id can be used to find the correct layer in the tileset. But now I go to bed)

nyurik:
ok, thx, i see the issue. I think the simplest solution would be to refactor the test file a bit to cover this usecase - I can do it myself tomorrow if its ok with you.  The rest of the code looks good except for one thing -- please add a property `requires`:

```
    @property
    @deprecated(version='5.4.0', reason='use requires_layers property instead')
    def requires(self) -> List[str]:
        return self.requires_layers
```

nyurik:
I reworked your tests a bit, please let me know if you are ok with the changes, and I can merge things once all tests pass (optionally we might want another person to review)

nyurik (approved):

nyurik:
nope, found another bug (fixed) -- `to_sql` gets called for each schema (SQL file) in a layer. We don't want to assert for each schema, we just want to do it once per layer.

tests/python/test_sql.py
```diff
@@ -121,6 +121,47 @@ def test_require(self):
...
                    dict(c1__c2__c4__c3=[c1, c2, c4r12, c3r2]))
 
+        self._test("a13", [c2, c7r2], dict(c2__c7=[c2, c7r2]))
+        self._test("a14", [c1, c2, c8r12], dict(c1__c2__c8=[c1, c2, c8r12]))
+        self._test("a15", [c9], dict(c9=[c9]))
+        self._test("a16", [c10], dict(c10=[c10]))
+        self._test("a17", [c11], dict(c11=[c11]))
+        self._test("a18", [c12], dict(c12=[c12]))
+
+    def _parse_reqs(self, reqs, expected_layers, expected_tables, expected_funcs):
```
Falke-Design, line 131: would be `_parse_props` not better instead `_parse_reqs` because you test here all properties not only requires

Falke-Design:
@nyurik  Nice work! Thank you for your help

nyurik:
@Falke-Design thx, updated. Could you update the first comment to describe the current design? I will copy that comment into the squash commit message when ready.

Falke-Design:
@nyurik Done. Is this ok so? Else can you please change it?

TomPohys:
Thank you both for working on this! It looks great!

TomPohys (approved):

Issue #363, opened 2021-08-29
Changed IMPOSM Repo makes problems with `make import-sql`


Falke-Design:
With the newest versions (from Github) of openmaptiles and openmaptiles-tools I get following error while calling `make import-sql`:

> psql:/sql/parallel/transportation__transportation_name.sql:232: ERROR:  there is no unique or exclusion constraint matching the ON CONFLICT specification

The problem part:
```
ALTER TABLE osm_route_member ADD COLUMN IF NOT EXISTS concurrency_index int;

INSERT INTO osm_route_member (id, concurrency_index)
  SELECT
    id,
    ROW_NUMBER() over (PARTITION BY member ORDER BY network_type, network, LENGTH(ref), ref) AS concurrency_index
  FROM osm_route_member
  ON CONFLICT (id) DO UPDATE SET concurrency_index = EXCLUDED.concurrency_index;
```

The sql file:
[transportation__transportation_name_error.sql.txt](https://github.com/openmaptiles/openmaptiles/files/7070921/transportation__transportation_name_error.sql.txt)

The problem comes with the commit 4211019a where in the Dockerfile the IMPOSM Repo is changed from a custom branch / fork to the offical one.

## Steps to reproduce:
Get the repos:
```
git clone https://github.com/openmaptiles/openmaptiles.git
git clone https://github.com/openmaptiles/openmaptiles-tools.git
```

Edit docker-compose.yml in `openmaptiles` to use the local build of openmaptiles-tools (more infos in: #361):
```
  openmaptiles-tools:
    #image: "openmaptiles/openmaptiles-tools:${TOOLS_VERSION}" #comment
    build:
      context: YOUR_PATH_TO/openmaptiles-tools
      dockerfile: Dockerfile
    env_file: .env
```

Call in `openmaptiles`:
```
make
make start-db
make import-data
make download area=liechtenstein
make import-osm
make import-borders
make import-wikidata
make import-sql <-- throws the error
```

Falke-Design:
@nyurik @ZeLonewolf @TomPohys can please check some one else this issue, this error is still happening with the newest version of  `openmaptiles-tools`

ZeLonewolf:
I believe that SQL code cited here is not the latest, can you confirm?

Falke-Design:
Yes you are right the sql changed but error is the same.

The current Error:
> *** ERROR detected, aborting:
[...]

Issue #361, opened 2021-08-28
Use local openmaptiles-tools repo for openmaptiles (Docker)


Falke-Design:
#### Connect local `openmaptiles-tools`
The service `openmaptiles-tools` needs to be changed in [`docker-compose.yml`](https://github.com/openmaptiles/openmaptiles/blob/master/docker-compose.yml#L30) in `openmaptiles`.
Replace `image` with `build`
```
  openmaptiles-tools:
    image: "openmaptiles/openmaptiles-tools:${TOOLS_VERSION}"
    env_file: .env
```
with
```
  openmaptiles-tools:
    #image: "openmaptiles/openmaptiles-tools:${TOOLS_VERSION}" #comment
    build:
      context: YOUR_PATH_TO/openmaptiles-tools
      dockerfile: Dockerfile
    env_file: .env
```
Now it uses the local path for the docker image.

#### Generate new docker image after changes:
For every change in `openmaptiles-tools` the docker image needs to be generated new:
For this remove the old docker image:
```
# list all docker images:
docker images

# search for the openmaptiles-tools image and get the `IMAGE ID`

# Remove the docker image
docker rmi IMAGE_ID
```

or simply use:
```
docker rmi $(docker images | grep openmaptiles-tools) #maybe there is a better command for this
```

And then you can simply call your `openmaptiles`-make-script again and the docker image will be generated automatically again.
For example: `make import-osm`

------

I'm a newbe in openmaptiles and in docker so I had some trouble to find out how I can manage this. Because of that I created this "issue" to document it for other peoples. Maybe this is not the cleanest or best way but it a working solution. If someone has a better solution please let me know.

To the Maintainers: Maybe some of you want to add this to the Readme (with a better english spelling) or you can simply close this issue.

TomPohys:
Thanks, it looks nice. The readme could be updated. 

Another solution could be:

1. Change the line in [`openmaptiles/__init__.py`](https://github.com/openmaptiles/openmaptiles-tools/blob/master/openmaptiles/__init__.py#L1) and replace it with e.g. 

```diff
- __version__ = "5.3.2"
+ __version__ = "my_test"
```
2. build your local dockers.
3. In the OpenMapTiles repository in [`.env` file](https://github.com/openmaptiles/openmaptiles/blob/master/.env#L7) replace

```diff
- TOOLS_VERSION=5.3
+ TOOLS_VERSION=my_test
```

But I think both solutions could work.

[...]

PR #373, opened 2021-09-19
Add helpText to requires

Falke-Design:
Add `helpText` to `layer.requires` like discussed in https://github.com/openmaptiles/openmaptiles/issues/1220

I had to change from `SELECT 'osm_ocean_polygon'::regclass;` to `PERFORM 'osm_ocean_polygon'::regclass;` because plpgsql needs a return value if `SELECT` is used.

Also added an error if a function has no arguments:
```
when invalid_text_representation then
        RAISE EXCEPTION '%! The arguments of the required function "osmFunc" of the layer "water" are missing. Example: "osmFunc(text)"', SQLERRM;
```

Example:
```
layer:
  id: "water"
  requires:
    helpText: 'This is a text with "quotes"'
    functions: "osmFunc"
    tables: "osm_ocean_polygon"
```

SQL:
```
-- Assert osm_ocean_polygon exists
do $$
begin
        PERFORM 'osm_ocean_polygon'::regclass;
exception when undefined_table then
        RAISE EXCEPTION '%! This is error text with "quotes"', SQLERRM;
end;
$$ language 'plpgsql';

-- Assert osmFunc exists
do $$
begin
        PERFORM 'osmFunc'::regprocedure;
exception when undefined_function then
        RAISE EXCEPTION '%! This is a text with "quotes"', SQLERRM;
when invalid_text_representation then
        RAISE EXCEPTION '%! The arguments of the required function "osmFunc" of the layer "water" are missing. Example: "osmFunc(text)"', SQLERRM;
end;
$$ language 'plpgsql';
```

Result of `make import-sql`:
> psql:/sql/parallel/water__waterway.sql:10: ERROR:  relation "osm_ocean_polygon2" does not exist! This is a text with "quotes"


Falke-Design:
@nyurik thanks for cleaning up! I didn't used `DETAIL` or `HINT` because now the Exception Message is not anymore in one line and can't be catched anymore as "full stacktrace" and printed again at the end.

```
DO
Time: 0.871 ms
psql:/sql/parallel/water__waterway.sql:14: ERROR:  expected a left parenthesis
DETAIL:  Required function "osmFunc" in layer "water" is incorrectly declared. Use full function signature with parameter types, e.g. "my_magic_func(TEXT, TEXT)"
CONTEXT:  PL/pgSQL function inline_code_block line 9 at RAISE
Time: 0.486 ms
xargs: sh: exited with status 255; aborting

*** ERROR detected, aborting:
psql:/sql/parallel/water__waterway.sql:14: ERROR:  expected a left parenthesis
make: *** [Makefile:422: import-sql] Error 1
```

If you want to throw the Exception with `DETAIL` then I suggest that I update the PR https://github.com/openmaptiles/openmaptiles/pull/1237
Then it should be like: If ERROR found, print ERROR Line. If ERROR and then in the next lines DETAIL found, print DETAIL

nyurik:
Oh, i didn't think about that...  do you think it would be possible to print last N surrounding lines instead of just the one line with `awk` / `grep` ?  Also, note that it is not just ERROR and DETAIL, it could also be HINT.  

We could of course make it into a single line, but it won't be as readable i think?

Falke-Design:
I found an way, but now the DETAIL text is printed before the ERROR text:
grafik

```
awk '1{print; fflush()} $$0~".*ERROR" {err[NR]=$$0; errFound=1;} $$0~".*DETAIL" { if(errFound == 1){err[NR]=$$0;} }  END{ if(length(err) > 0){print "\n*** ERROR detected, aborting:"; for(i in err){print err[i]}; exit(1)} }'
```

Simplified explanation:

- Check if DETAIL or ERROR is found in line
- if ERROR found: line is added to the key-value array `err`. The key is the line number `NR` `err[NR]=$$0`. And the flag errFound is set to `1`
- if DETAIL found and flag errFound is `1`: line is added to the key-value array `err`.
- At the end of the output it loops through the error array.

nyurik:
I suspect it will be far simpler to edit the [import-sql](https://github.com/openmaptiles/openmaptiles-tools/blob/master/bin/import-sql) -- and do all the error detection there. This way we can even capture things with `tee` and `grep` the output to just the lines around the error (for example)

nyurik:
P.S. i think once we update the `import-sql` script, we can release a new tools version - long overdue.  Let me know how you want to proceed with this.

Falke-Design:
It would be much cleaner to do this in `import-sql` but that was not possible for me. Because of the parallel processing with `find` and `xarg` I was not able to catch the errors and then print the errors  again.
When a error happens while a sql process, it "informs" xarg to stop all other parallel processes, xarg run `exit 255`. After this exit I was not able to run code to print the error again. 

The result was that i switched to the Makefile with `awk`

Falke-Design:
To the release:

if it is ok for you i will add in the next 2 days a documention about #361 to the `/docs`.

And I have to test again if this error still happens: #363 (maybe you can test it too)

nyurik:
I think I will do #376 - rewrite `import-sql` in python, which will have proper error output

PR #383, opened 2021-10-22
Implement quick tile generation with imputing

nyurik:
This rewrites generate-tiles script:

If run without MID_ZOOM parameter, this script will generate all tiles from MIN_ZOOM to MAX_ZOOM
If MID_ZOOM is set, it will first generate all tiles from MIN_ZOOM to MID_ZOOM. Afterwards,
the script will generate one zoom level at a time, making sure to only generate non-empty tiles.
A non-empty tile is a tile that has some data in the previous zoom. All other tiles will be imputed
using "mbtiles-tools impute" command.

Additionally, the script will automatically detect tileset min-zoom and max-zoom from the openmaptiles.yaml file, and will run mbtiles-tools meta-generate at the end.

Lastly, this also changes the version to 0.0.0 to avoid accidental publishing (the workflow will automatically fix this when publishing)


Falke-Design:
Can you explain me for what this is needed? I don't understand it 🙃

Issue #366, opened 2021-09-07
Build fail for import-data docker image


ache051:
Build fail for the last few days due to Natural Earth's download service being down or removed: https://www.naturalearthdata.com/downloads/. With the download links on Natural Earth's website striked-thru, it can be certain that the NE team is either aware of the issue or the change was deliberate.
Issue is also submitted in NE's repo: https://github.com/nvkelso/natural-earth-vector/issues/577

Suggestion for fix:
Natural Earth data is curated on AWS' Open Data Registry: https://registry.opendata.aws/naturalearth/. The issue can be fixed if line 11 of the Dockerfile:
https://github.com/openmaptiles/openmaptiles-tools/blob/4dfdc1ab1ccdb0475cd94ebd8570fc255e3f8024/docker/import-data/Dockerfile#L11
is changed to:
wget --quiet https://naturalearth.s3.amazonaws.com/packages/natural_earth_vector.sqlite.zip  ;\

nyurik:
i'm all for migrating to that link (thx!)

nyurik:
@ache051 can you create a PR for this?  Don't want to take credit for your research :)

ache051:
Done :)

Issue #354, opened 2021-08-01
Updating lake centerlines


ache051:
Hi,
The current OMT vectortiles generation routine is still [pulling data](https://github.com/openmaptiles/openmaptiles-tools/blob/3796c4eab03dd8a317cea2d2c9b0c7dbdce35ac9/docker/import-data/Dockerfile#L26-L31) from Lukas' [osm-lakeline](https://github.com/lukasmartinelli/osm-lakelines) released in 2016. Granted lake centerlines don't change much but would still good to update periodically. At the moment I forked Lukas' repository and changed a few lines so centerlines ae generated when I run quickstart.sh from OpenMapTiles. Not pretty but does the job.

My question is if there is any plan on either updating the data or incorporating functionality from osm-lakeline into openmaptiles-tools or vectortiles generation routine?

TomPohys:
Thanks. Yes, there is a plan to figure it out. Do you have some measured time, how long it takes to regenerated osm-lakeline on your use case?

ache051:
I have started doing a run from a planet.osm.pbf-loaded database on AWS (m5.24xlarge, 96 cores) and also on local machine (AMD Ryzen9 3950x, 12 cores) to get some numbers. Will post them when they are done.

ache051:
The AWS instance finished generating the centerlines (36370 of them in total) in 132 minutes. Will check the local machine tomorrow.

ache051:
The local machine finished in 155 minutes with 37544 centerlines (planet.osm.pbf is 2 months newer than AWS instance's).
For reference the AWS instance needed 8 days to run quickstart.sh down to zoom level 14, and the local machine needed 31 days.

nyurik:
@ache051 ideally we should NOT use the lakeline centerlines like we do now.  My long term hope was to get rid of it entirely, and instead use the usual PostGIS functions to compute it, possibly generating a materialized view if needed for performance reasons. The current implementation adds significant complexity to the project's ecosystem.

ache051:
@nyurik I agree with you 100%. Exporting to shapefile then generate lines is cumbersome. At the time being though it is a stop gap solution for my project. If anybody thinks there is merit, then I will look at if [ST_ApproximateMedialAxis](https://postgis.net/docs/ST_ApproximateMedialAxis.html) can do the trick.
I have posted the centerlines generated from [last night's run](https://github.com/ache051/osm-lakelines/releases/tag/v1.0) for your amusement.

[...]

PR #390, opened 2021-11-03
Update lake centerlines

lazaa32:
This PR does few things:

1. Lake centerlines were re-generated from latest planet (planet-211018.osm.pbf). There are now 38768 centerlines, instead of 14790. 
2. Centerlines in geojson are already in 3857 so there is no re-projection needed during import.
3. Links to lake_centerline.geojson and natural_earth_vector.sqlite.zip were updated to prevent a situation when data will be removed (https://github.com/openmaptiles/openmaptiles-tools/issues/366).
4. I kept `water-polygons-split-3857.zip` link as these data are regularly updated.

### DIFF
--- a/docker/import-data/Dockerfile
+++ b/docker/import-data/Dockerfile
@@ -8,7 +8,7 @@ RUN set -eux  ;\
     mkdir -p "$DIR"  ;\
     cd "$DIR"  ;\
     apk add --no-cache sqlite  ;\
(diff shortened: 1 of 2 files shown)


TomPohys (approved):

nyurik:
really can't wait for us to switch away from the lakeline dump generation. It's ridiculous that we cannot construct the same data directly in PostGIS.  See #354

PR #399, opened 2022-01-27
Implement buffer_size overrides

nyurik:
See https://github.com/openmaptiles/openmaptiles/issues/1345

Layer file changes:
* Added support for an optional `min_buffer_size` value, which is only used in case of a global override, but if set, it must be less than or equal to `buffer_size`.
* Layer's `buffer_size` is now optional if `min_buffer_size` is set.

Tileset file changes:
* Added support for an optional `overrides` section (similar to `defaults`).
* Currently only suppports optional `overrides.buffer_size` int value. If set, this value will override the `buffer_size` set in the layer.  If layer has `min_buffer_size`, the largest of the two will be used.
* An environment variable `TILE_BUFFER_SIZE` can be used instead of overrides.buffer_size, and takes precedent.

### DIFF
--- a/openmaptiles/tileset.py
+++ b/openmaptiles/tileset.py
@@ -1,6 +1,7 @@
 from dataclasses import dataclass
+from os import getenv
 from pathlib import Path
-from typing import List, Union, Dict, Any, Callable
+from typing import List, Union, Dict, Any, Callable, Optional
 
 import sys
 import warnings
@@ -181,7 +182,30 @@ class Layer:
 
     @property
     def buffer_size(self) -> int:
-        return self.definition['layer']['buffer_size']
+        """Each layer must have a default buffer size, with either `buffer_size` or `min_buffer_size` or both.
+        If both are set, buffer_size must be >= min_buffer_size.
+        min_buffer_size is only used when there is a global buffer size override,
+        e.g. if global is set to 0, and layer's min_buffer_size is set to 4, the result is 4.
+        """
+        size = self.definition['layer'].get('buffer_size')
+        min_size = self.definition['layer'].get('min_buffer_size')
+        if size is None and min_size is None:
+            raise ValueError(f'Layer "{self.id}" is missing an integer buffer_size and/or min_buffer_size')
+        elif size is not None and min_size is not None:
+            if size < min_size:
+                raise ValueError(f'Layer "{self.id}" has buffer_size less than min_buffer_size')
+        elif size is None:
+            # size is not set, use min_size as default
+            size = min_size
+        else:
+            # size is set, min_size is not set
+            min_size = 0
+
+        override = self.tileset.override_buffer_size if self.tileset else None
+        if override is not None:
+            size = max(override, min_size)
+
+        return size
 
     @property
     def max_size(self) -> int:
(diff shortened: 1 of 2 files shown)


farfromrefug:
@nyurik this look pretty awesome!

openmaptiles/tileset.py
```diff
@@ -181,7 +196,49 @@ def description(self) -> str:
 
     @property
     def buffer_size(self) -> int:
-        return self.definition['layer']['buffer_size']
+        """Each layer must have a default buffer size, with either `buffer_size` or `min_buffer_size` or both.
+        If both are set, buffer_size must be >= min_buffer_size.
+        min_buffer_size is only used when there is a global buffer size override,
```
zstadler, line 201:
```suggestion
        min_buffer_size is only used when there is an external buffer size override,
```

The override can be either global or layer-specific

openmaptiles/tileset.py
```diff
@@ -181,7 +196,49 @@ def description(self) -> str:
 
     @property
     def buffer_size(self) -> int:
-        return self.definition['layer']['buffer_size']
+        """Each layer must have a default buffer size, with either `buffer_size` or `min_buffer_size` or both.
+        If both are set, buffer_size must be >= min_buffer_size.
```
zstadler, line 200:
```suggestion
        If both are set, `buffer_size` must be >= `min_buffer_size`.
```

openmaptiles/tileset.py
zstadler, line 201: My understanding is that `min_buffer_size` is also used with per-layer override.

```suggestion
        `min_buffer_size` is only used when there is a `buffer_size` override.
```

openmaptiles/tileset.py
```diff
@@ -181,7 +196,49 @@ def description(self) -> str:
...
+        min_buffer_size is only used when there is a global buffer size override,
+        e.g. if global is set to 0, and layer's min_buffer_size is set to 4, the result is 4.
```
zstadler, line 202:
```suggestion
        e.g. if global `buffer_size` is set to 0, and layer's `min_buffer_size` is set to 4, the result is 4.
```

openmaptiles/tileset.py
```diff
@@ -181,7 +196,49 @@ def description(self) -> str:
...
+        Per layer tileset overrides are allowed for both buffer_size and min_buffer_size.
```
zstadler, line 203:
```suggestion
        Per layer tileset overrides are allowed for both `buffer_size` and `min_buffer_size`.
```

openmaptiles/tileset.py
```diff
@@ -181,7 +196,49 @@ def description(self) -> str:
...
+        Per layer overrides have higher priority than global overrides, but less than ENV var.
```
zstadler, line 204: My understanding of this sentence is that env. variable has higher priority than per-layer min_buffer_size override. 

```suggestion
        Per layer `buffer_size` override has a higher priority than a global override, but less than ENV var.
        Per layer `min_buffer_size` tileset override has a higher priority than the layer's `min_buffer_size`.
        The resulting `min_buffer_size` is used if it is higher that the resulting `buffer_size`.
```

As far as I understand the code, the final buffer size of a layer is
```
max(
  first_of(
    TILE_BUFFER_SIZE environment variable,
    layer-specific override of the `buffer_size`,
    global override of the `buffer_size`,
    layer's `buffer_size`,
    layer's min_buffer_size,
    0),
  first_of(
    layer-specific override of min_buffer_size,
    global override of min_buffer_size,
    layer's min_buffer_size )
)
```

nyurik (commented):
tried to cleanup buffer size logic/docs, thx for feedback!

TomPohys (approved):

PR #401, opened 2022-01-31
Add generate-tiles LIST_FILE support

nyurik:
Allow generate-tiles to generate just the tiles given in a file.
This allows the same functionality as [export-list.sh](https://github.com/openmaptiles/openmaptiles-tools/blob/207dbb48e38b60eb7037951b19857d01e3c31b89/docker/generate-vectortiles/export-list.sh) using modern ST_AsMVT approach.


README.md
```diff
@@ -353,7 +353,9 @@ generate-doc <tileset>
 ### Generate MBTiles tile snapshot file
 
 `generate-tiles` script generate tiles from PostgreSQL using [tilelive-copy](https://github.com/mapbox/tilelive#bintilelive-copy) with [tilelive-pgquery](https://github.com/nyurik/tilelive-pgquery#tilelive-pgquery).
-If run without `MID_ZOOM` environment variable, this script will generate all tiles from `MIN_ZOOM` to `MAX_ZOOM` (inclusive).
+If run with `LIST_FILE` environment variable, this script will only generate tiles given in the list file, and it will ignore min/mid/max zoom values. For example, set `LIST_FILE=/export/tiles.txt` to match what was previously done with `export-list.sh` script.
+
+If run without `LIST_FILE/MID_ZOOM` environment variables, this script will generate all tiles from `MIN_ZOOM` to `MAX_ZOOM` (inclusive).
```
zstadler, line 358:
```suggestion
If the `LIST_FILE` and `MID_ZOOM` environment variables are not set, this script will generate all tiles from `MIN_ZOOM` to `MAX_ZOOM` (inclusive).
```

bin/generate-tiles
```diff
@@ -6,7 +6,9 @@ set -o nounset
 #
 # Generate tiles from postgres using tilelive-copy and tilelive-pgquery
 #
-# If run without MID_ZOOM parameter, this script will generate all tiles from MIN_ZOOM to MAX_ZOOM
+# If run with LIST_FILE parameter, this script will only generate tiles given in the list file,
```
zstadler, line 9:
```suggestion
# If the LIST_FILE environment variable is set, this script will only generate tiles given in the list file,
```

bin/generate-tiles
```diff
@@ -6,7 +6,9 @@ set -o nounset
...
+# and it will ignore min/mid/max zoom values. For example, set LIST_FILE=/export/tiles.txt .
+# Otherwise, if run without MID_ZOOM parameter, generate all tiles from MIN_ZOOM to MAX_ZOOM.
```
zstadler, line 11:
```suggestion
# Otherwise, if the MID_ZOOM environment variable is not set, generate all tiles from MIN_ZOOM to MAX_ZOOM.
```

bin/generate-tiles
```diff
@@ -84,7 +86,13 @@ function run_tilelive_copy() {
 }
 
 
-if [[ -z "${MID_ZOOM-}" ]]; then
+if [[ -n "${LIST_FILE-}" ]]; then
+
+  # Generate all tiles given in a file LIST_FILE.
+  echo "$(date '+%Y-%m-%d %H-%M-%S') Generating tiles from a list $LIST_FILE from $HOST_COUNT servers, using $MAX_HOST_CONNECTIONS connections per server, $ALL_STREAMS streams"
+  run_tilelive_copy --scheme=list "--list=$LIST_FILE" --timeout="${TIMEOUT:-1800000}"
```
zstadler, line 93:
```suggestion
  run_tilelive_copy --scheme=list --list="$LIST_FILE" --timeout="${TIMEOUT:-1800000}"
```

zstadler: In addition, 

1. it may be simpler to place `--timeout` within the `run_tilelive_copy` function rather than adding `--timeout="${TIMEOUT:-1800000}"` whenever it is called.
2. It may be more appropriate for the default `TIMEOUT` to be handled together with the other defaults at the beginning of the script

nyurik: the timeout default is different in the mid-zoom handling section, so it won't be as straightforward to send two different defaults into mid-zoom

zstadler:
Looks good!!!

PR #402, opened 2022-02-04
Add vars on layers (with overrides)

francois2metz:
I wanted to make some stuff configurable via environnement variable. So I added environnement variable substitution when collecting sql files.

### DIFF
--- a/openmaptiles/sql.py
+++ b/openmaptiles/sql.py
@@ -253,22 +253,24 @@ def to_sql(sql: str, layer: Layer, nodata: bool):
     """Clean up SQL, and perform any needed code injections"""
     sql = sql.strip()
 
-    # Replace '%%FIELD_MAPPING: <field_name>%%' with fields from layer definition
-    def field_map(match):
-        return FieldExpander(match.group(2), layer, match.group(1)).parse()
+    # Substitute  "%% <cmd> : <param> %%"   with cmd-specific value:
+    #   * FIELD_MAPPING: field_name  -> generated SQL CASE statement
+    #   * VAR: var_name              -> a variable value from the layer
+    def substitute(match):
+        indent = match.group(1)
+        cmd = match.group(2)
+        param = match.group(3)
+        if cmd == 'FIELD_MAPPING':
+            return FieldExpander(param, layer, indent).parse()
+        elif cmd == 'VAR':
+            result = layer.get_var(param)
+            if result is None:
+                raise ValueError(f'Variable {param} does not exist in layer {layer.id}')
+            return indent + result
+        else:
+            raise ValueError(f'Unrecognized substitution {cmd}')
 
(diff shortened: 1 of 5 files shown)


nyurik:
Thanks @francois2metz , this does look like an interesting idea!  I think we should not limit it to `ENV` because env variables is an implementation detail that should not be hard coded.  Instead, I think we should introduce a concept of a "variable", and have multiple ways to set that variable. Something like this:

#### In SQL file
```sql
# layers/water/water.sql
SELECT * from test where zoom > %% VAR: custom_zoom %%
```
#### In Layer yaml file
```yaml
# layers/water/water.yaml
layer:
  vars:
    custom_zoom: 10
```
#### In Tileset yaml file
```yaml
# openmaptiles.yaml
tileset:
  layers:
    - file: layers/water/water.yaml
      vars:
        custom_zoom: 12
  overrides:
    vars:
      custom_zoom: 11
```
#### ENV vars
`OMT_VAR_custom_zoom=13`

The above would work as following:
* The layer defines and uses `custom_zoom` variable. It is an error to have `%% VAR: foo %%` if `foo` default value is not set in the layer's yaml file
* Tileset can override the var globally (i.e. `10 -> 11`)
* Tileset can override the var per layer - in the example above, global override is lower priority than per-layer override, so it becomes `11 -> 12`
* Lastly the ENV var can be used to override all of the above, setting `custom_zoom` to 13.  This approach is the same as used by [Terraform](https://www.terraform.io/language/values/variables#environment-variables). Note that env var names are case-sensitive on Linux & Mac, but not on Windows.

francois2metz:
Sound like a good idea. I'll work on it!

nyurik:
@francois2metz i converted it to draft, please convert it back when ready. Thx!

nyurik:
@francois2metz I think it would be easier to implement all this logic in the layer init function. Basic logic:
* create a dict of vars from the layer yaml
* get global tileset vars, and iterate over all of them.  Only override layer if it already had that var
*same for layer overrides, but raise an error if layer doesn't already define that var
* iterate over resulting dict, and try to get corresponding env vars
* store the resulting dict in self, and create a getter function

francois2metz:
@nyurik Hi, I think I have done most of the work. I would be great to have a new review!

francois2metz:
I updated the PR on indoor= that use this code: https://github.com/indoorequal/indoorequal/pull/30

nyurik:
@francois2metz I have made a few proposed changes to your code -- see https://github.com/indoorequal/openmaptiles-tools/pull/1

francois2metz:
I manually merged your PR. The only change is the name of the TestCase `TilesetTestCase` that match the name of the tested class.

nyurik (approved):
looks good, thanks!

PR #404, opened 2022-02-21
Update README.md

zstadler:
Explain how to build the docker images.

### DIFF
--- a/README.md
+++ b/README.md
@@ -5,6 +5,7 @@ It includes tools to prepare Imposm mappings and SQL files based on layers defin
 We encourage other people to use this for their vector tile projects as well since this approach works well for us.
 
 ## Docker Images
+The repository hosts the code for the following Docker images. They can be built using `make build-all-dockers`.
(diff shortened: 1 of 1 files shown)


TomPohys (approved):

TomPohys:
Thanks!

zstadler:
@TomPohys,

It looks like this PR is incomplete for a couple of reasons.

First, all docker image versions are '0.0.0'. This can be changed using `make VERSION=latest build-all-dockers`
Second, when building the `latest` like that, the python programs still report a `0.0.0` version.

For example:

```
$ docker run --rm openmaptiles/openmaptiles-tools:latest generate-doc --version
0.0.0
```

What should README.md say about fixing the second issue?

TomPohys:
Thanks for pointing this out. There are a lot of scripts that are using simple `openmaptiles.__version__` [(all in search)](https://github.com/openmaptiles/openmaptiles-tools/search?q=__version__&type=code)

or from your example
https://github.com/openmaptiles/openmaptiles-tools/blob/b8f0ae329f6e295a0abad741e0345e6a4abbaa85/bin/generate-doc#L32

One option is to change the value in the file `openmaptiles/init.py` directly.

zstadler:
@TomPohys,

Perhaps [`openmaptiles/__init__.py`](https://github.com/openmaptiles/openmaptiles-tools/blob/master/openmaptiles/__init__.py) should be
```
__version__ = 'latest'
```
and the version will continue be overwritten in the release branches.

If you're OK with this, I'd be happy to send a PR for both corrections.

PR #406, opened 2022-02-23
Add $BBOX info to generate-tiles log

zstadler:
Clarifies the extend used for generating the tiles


zstadler:
Superseded by  #409

Issue #313, opened 2020-09-04
Implement download-diff


zstadler:
[`import-update`](https://github.com/openmaptiles/openmaptiles-tools/blob/master/bin/import-update) repeatedly pulls the changes from the replication server and applies them according the `replication_interval` defined in the `${IMPOSM_CONFIG_FILE}` file. It works well for the daily updates available for Planet and Geofabrik. It is not applicable for minutely updates because it cannot be active in parallel with the required refresh of the mbtiles by [`export-list.sh`](https://github.com/openmaptiles/openmaptiles-tools/blob/master/docker/generate-vectortiles/export-list.sh). This limitation is caused by a PostgreSQL restriction preventing read access to materialized views while they are being updated.

Therefore, a continuous minutely update of mbtiles would run [`export-list.sh`](https://github.com/openmaptiles/openmaptiles-tools/blob/master/docker/generate-vectortiles/export-list.sh) in-between subsequent runs of [`import-diff`](https://github.com/openmaptiles/openmaptiles-tools/blob/master/bin/import-diff). AFAIK, there is no tool in `openmaptiles-tools` to provide the required input for [`import-diff`](https://github.com/openmaptiles/openmaptiles-tools/blob/master/bin/import-diff).

The `download-diff` would use the replication URL created by `download-osm` in the `${IMPOSM_CONFIG_FILE}` file to:
- Download the required updates from the replication, according to the latest timestamp in `${area}.pbf`
- Assemble them into the `changes.osc.gz` file as input for required by [`import-diff`](https://github.com/openmaptiles/openmaptiles-tools/blob/master/bin/import-diff)
- Update the `${area}.pbf` file

The minutely updates flow would repeat the following sequence:
1. Execute `download-diff` to combine all new minutely updates into the `changes.osc.gz` file.
2. Execute [`import-diff`](https://github.com/openmaptiles/openmaptiles-tools/blob/master/bin/import-diff) to update the PostGIS DB and create the list of expired tiles.
3. Extend the list of expired tiles to all zoom levels between `MIN_ZOOM` and `MAX_ZOOM`, similar to @frodrigo's [tile_multiplier.py](https://github.com/makinacorpus/makina-maps/blob/master/nginx/tile_multiplier.py) 
4. Execute [`export-list.sh`](https://github.com/openmaptiles/openmaptiles-tools/blob/master/docker/generate-vectortiles/export-list.sh) to update the expired tiles in the mbtiles file.

[...]

Issue #290, opened 2020-06-03
Updating tiles other than zoom=14


ingarpedersen:
I'm a bit confused about the diff update process. 
I have a database up and running, and I'm running a diff update. This update results in a file that lists all tiles that is affected by the update, but only at zoom=14. How do I get a list of all tiles that needs to be updated at all zoom levels? 
I have verified that a change (a new road) was missing at zoom 12 but became visible on zoom=14. If run full tile generation the change is visible at zoom=12.

Thanks
Ingar F Pedersen

zstadler:
When running diff update, imposm creates a list of tiles, in zoom14, that need to be updated, while the tile generation accepts a list tiles to be generated in all relevant zoom levels.
The expansion of the list to all zoom levels between `MIN_ZOOM` and `MAX_ZOOM` can be done using @frodrigo's [tile_multiplier.py](https://github.com/makinacorpus/makina-maps/blob/master/nginx/tile_multiplier.py) 

The conversion could be done this way:

```
find ./data -name "*.tiles" -exec cat {} \; -exec rm {} \; | \
    tile_multiplier.py $(MIN_ZOOM) $(MAX_ZOOM) > data/tiles.txt
```

PR #407, opened 2022-02-25
Add `tile_miltiplier`

zstadler:
Expand a list of changed tiles to higher and/or lower zoom levels.

This program expands the list of tiles produced by [`import-diff`](https://github.com/openmaptiles/openmaptiles-tools/blob/master/bin/import-diff) to be used by [`generate-tiles`](https://github.com/openmaptiles/openmaptiles-tools/blob/master/bin/generate-tiles).
[`import-diff`](https://github.com/openmaptiles/openmaptiles-tools/blob/master/bin/import-diff) produces a list of tiles in zoom `MAX_ZOOM` that need updating.
When [`generate-tiles`](https://github.com/openmaptiles/openmaptiles-tools/blob/master/bin/generate-tiles) is used with `FILE_LIST`, it expects a list of tiles in all zoom levels.

- For each line of the form `z/x/y` in its standard input, representing a tile, the program outputs all tiles that overlap it for all zoom levels between `<min_zoom>` and `<max_zoom>`.
- Each tile will appear once in its output.
- The input zoom level need not be within the `<min_zoom>..<max_zoom>` range.

Thanks @frodrigo for https://github.com/makina-maps/makina-maps/blob/master/nginx/tile_multiplier.py

### DIFF
--- a/bin/tile_multiplier
+++ b/bin/tile_multiplier
@@ -1,0 +1,56 @@
+#!/usr/bin/env python3
+"""
+Expand a list of changed tiles to higher and/or lower zoom levels.
+
+For each line of the form z/x/y in its standard input, representing a tile, the program outputs all tiles that overlap it for all zoom levels between <min_zoom> and <max_zoom>.
+Each tile will appear once in its output.
+The input zoom level need not be within the <min_zoom>..<max_zoom> range.
+
+Usage:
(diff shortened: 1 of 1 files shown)


zstadler:
Related to #290, #313

TomPohys:
Thank you!

TomPohys (approved):

PR #416, opened 2022-09-21
Add style-tools binary

lazaa32:
This PR adds utility for handling style. Related to this PR in OMT: https://github.com/openmaptiles/openmaptiles/pull/1420

- It adds `spritezero` to generate sprite files from svg icons.
- It adds `style-tools` binary and `styleutils.py`. 

`style-tools split` to split `style.json` into style snippet `layer_name.json`, one per layer. Style snippets also keep the style layers `order` from the original style.json
`style-tools merge` to merge all style snippets into a new `style.json` with layers ordered based on `order` value.

### DIFF
--- a/bin/style-tools
+++ b/bin/style-tools
@@ -1,0 +1,40 @@
+#!/usr/bin/env python
+"""
+Usage:
+  style-tools split <tileset> <style>
+  style-tools merge <tileset> <style> <style_header>
+  style-tools recompose <tileset> <style> <style_header>
+  style-tools --help
+  style-tools --version
+
+Actions:
+  split           Break down style.json into separate jsons per yaml layer.
+  merge           Combine separate jsons into style.json.
+
+Options:
+  --help              Show this screen.
+  --version           Show version.
+"""
+from pathlib import Path
+from docopt import docopt
+import openmaptiles
+from openmaptiles.utils import print_err
+from openmaptiles.styleutils import merge, split
+
+if __name__ == '__main__':
+    args = docopt(__doc__, version=openmaptiles.__version__)
+    tileset = args['<tileset>']
+    tileset_fp = Path(tileset)
+    style = args['<style>']
+    style_fp = Path(style)
+    if args['split']:
+        split(tileset_fp, style_fp)
+    elif args['merge']:
+        style_header = args['<style_header>']
+        style_header_fp = Path(style_header)
+        merge(tileset_fp, style_fp, style_header_fp)
+    elif args['recompose']:
+        style_header = args['<style_header>']
+        style_header_fp = Path(style_header)
+        merge(tileset_fp, style_fp, style_header_fp)
+        split(tileset_fp, style_fp)
--- a/bin/style-tools
+++ b/bin/style-tools
@@ -20,10 +20,10 @@ from docopt import docopt
 import openmaptiles
 from openmaptiles.styleutils import merge, split
 
-if __name__ == '__main__':
+
+def main():
     args = docopt(__doc__, version=openmaptiles.__version__)
-    tileset = args['<tileset>']
-    tileset_fp = Path(tileset)
+    tileset_fp = Path(args['<tileset>'])
     style = args['<style>']
     style_fp = Path(style)
     if args['split']:
@@ -37,3 +37,7 @@ if __name__ == '__main__':
         style_header_fp = Path(style_header)
         merge(tileset_fp, style_fp, style_header_fp)
         split(tileset_fp, style_fp)
+
+
+if __name__ == '__main__':
+    main()
--- a/bin/style-tools
+++ b/bin/style-tools
@@ -18,7 +18,6 @@ Options:
 from pathlib import Path
 from docopt import docopt
 import openmaptiles
-from openmaptiles.utils import print_err
 from openmaptiles.styleutils import merge, split
 
 if __name__ == '__main__':
--- a/openmaptiles/styleutils.py
+++ b/openmaptiles/styleutils.py
@@ -1,4 +1,3 @@
-import yaml
 import json
 from pathlib import Path
 from openmaptiles.tileset import Tileset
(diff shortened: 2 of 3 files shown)


bin/style-tools
```diff
@@ -0,0 +1,40 @@
...
+from pathlib import Path
+from docopt import docopt
+import openmaptiles
+from openmaptiles.utils import print_err
+from openmaptiles.styleutils import merge, split
+
+if __name__ == '__main__':
+    args = docopt(__doc__, version=openmaptiles.__version__)
+    tileset = args['<tileset>']
+    tileset_fp = Path(tileset)
```
nyurik, line 26: This is shorter and easier to read (for this and other lines)

```suggestion
    tileset = Path(args['<tileset>'])
```

bin/style-tools
```diff
@@ -0,0 +1,40 @@
...
+  --help              Show this screen.
+  --version           Show version.
+"""
+from pathlib import Path
+from docopt import docopt
+import openmaptiles
+from openmaptiles.utils import print_err
+from openmaptiles.styleutils import merge, split
+
+if __name__ == '__main__':
```
nyurik, line 23: usually you don't want to have a real code inside the `if __name__` statement. Instead, move all the code to the `def main():`, and call that function from the if __name__ ... statement.

openmaptiles/styleutils.py
```diff
@@ -0,0 +1,87 @@
+import yaml
+import json
+from pathlib import Path
+from openmaptiles.tileset import Tileset
+
+
+def fp_to_dict(fp: str) -> dict:
```
nyurik, line 6:
```suggestion
def fp_to_dict(fp: Path) -> dict:
```

openmaptiles/styleutils.py
```diff
@@ -0,0 +1,87 @@
...
+    with open(fp, 'r') as f:
+        d = json.load(f)
+    return d
```
nyurik, line 9:
```suggestion
    return json.loads(fp.read_text(encoding="utf-8"))
```

openmaptiles/styleutils.py
```diff
@@ -0,0 +1,87 @@
...
+        ts_lyr_style_json_fp = get_ts_lyr_style_json_fp(ts_lyr)
+        ts_lyr_style_lyrs = list()
+        for style_lyr in style_lyrs_w_order:
+            source_layer = style_lyr.get('source-layer')
+            if source_layer == ts_lyr.filename.stem:
+                ts_lyr_style_lyrs.append(style_lyr)
+        out_dict = {'layers': ts_lyr_style_lyrs}
+
+        # write style layers with order to json for specific tileset layer
+        with open(ts_lyr_style_json_fp, 'w') as fout:
```
nyurik, line 50: can also use `ts_lyr_style_json_fp.write_text(json.dumps(...))`
(this var name is ... complicated):

openmaptiles/styleutils.py
```diff
@@ -0,0 +1,86 @@
...
+        lyr = lyrs[i]
+        lyr['order'] = i
+    return lyrs
+
+
+def get_order(layer: dict) -> int:
+    return int(layer.get('order'))
+
+
+def split(tileset_fp: Path, style_fp: Path):
```
nyurik, line 31: Also, i think it is much better to work with the style by its name, not path.  You can just do a `tileset.layers[style]` (and catch an error if its not there)

openmaptiles/styleutils.py
```diff
@@ -0,0 +1,86 @@
...
+    json_name = Path(yaml_path.name).with_suffix('.json')
+    ts_lyr_dir = yaml_path.parent
+
+    return ts_lyr_dir.joinpath(json_name)
+
+
+def add_order(lyrs: list) -> list:
+    for i in range(0, len(lyrs)):
+        lyr = lyrs[i]
+        lyr['order'] = i
```
nyurik, line 23: you want to use enumerate:

```suggestion
    for i, lyr in enumerate(lyrs):
        lyr['order'] = i
```

openmaptiles/styleutils.py
```diff
@@ -0,0 +1,86 @@
...
+    return lyrs
+
+
+def get_order(layer: dict) -> int:
+    return int(layer.get('order'))
```
nyurik, line 28: don't use `.get` here, because if 'order' is not there, `get` by itself will return `None`, and `int(None)` is an error

openmaptiles/styleutils.py
```diff
@@ -0,0 +1,86 @@
...
+
+
+def split(tileset_fp: Path, style_fp: Path):
+    tileset = Tileset.parse(tileset_fp)
+    tileset_lyrs = tileset.layers
+
+    # Add order to each style layer (e.g. backgroud - 0, labels - 196)
+    style = fp_to_dict(style_fp)
+    style_lyrs_w_order = add_order(style.get('layers'))
```
nyurik, line 37: same - add_order should not get a None value if style has no layers.

openmaptiles/styleutils.py
lazaa32: Applied.

bin/style-tools
lazaa32: Moved to `main()` function.

openmaptiles/styleutils.py
lazaa32: Renamed to `snippet_fp` and `snippet_lyrs`, I hope that's easier to understand.

openmaptiles/styleutils.py
lazaa32: Applied.

openmaptiles/styleutils.py
lazaa32: I added a check whether `order` is None and an error message.

openmaptiles/styleutils.py
lazaa32: I added a check whether `order` is None and an error message.

openmaptiles/styleutils.py
lazaa32: Applied.

TomPohys (approved):

PR #426, opened 2023-02-28
Add LabelCenterline function

### LINKED ISSUE #354: Updating lake centerlines

quincylvania:
Closes #354. The goal is to replace [acalcutt/osm-lakelines](https://github.com/acalcutt/osm-lakelines) with a faster SQL function. I've never written anything this complex in SQL before, so any help with efficiency and best practices is appreciated. I'm also new to this project, so I'm not sure about where to hook this in.

The logic:
1. Get the exterior polygon (ignoring islands) and use a convex hull if for some reason there's more than one exterior polygon
2. Use ST_LineInterpolatePoints to evenly distribute vertices around the polygon
3. Get ST_VoronoiLines from the interpolated vertices and covert to edges
4. Filter to only edges within the polygon
5. Recursively clip the shortest branch edges until only a single line is left
6. Simplify, smooth, and return

SFCGAL is not required for this method.

It takes about 40 seconds to calculate lines for the 395 largest lakes of Michigan, which I think is already a big improvement over osm-lakelines. The recursion takes about 75% of this time and can probably be made vastly more efficient.

The results are relatively consistent with osm-lakelines and I think look pretty good, except when there are large islands or bays.

Screenshot 2023-02-27 at 9 47 15 PM
Screenshot 2023-02-27 at 9 47 01 PM
Screenshot 2023-02-27 at 9 45 37 PM

### DIFF
--- a/sql/LabelCenterline.sql
+++ b/sql/LabelCenterline.sql
@@ -1,0 +1,69 @@
+/******************************************************************************
+### LabelCenterline ###
+
+Given a polygon or multipolygon, calculates a linestring suitable for a label.
+
+__Parameters:__
+
+- `geometry` input - A polygon or multipolygon.
+
+__Returns:__ `geometry(linestring)`
+******************************************************************************/
+CREATE OR REPLACE FUNCTION CountDisconnectedEndpoints(polyline geometry, line geometry) RETURNS integer
+    AS 'SELECT ST_NPoints(ST_RemoveRepeatedPoints(ST_Points(polyline)))
+                    - ST_NPoints(ST_RemoveRepeatedPoints(ST_Points(ST_Difference(polyline, line))))
+                    - ST_NPoints(line) + 2;'
+    LANGUAGE SQL
+    IMMUTABLE
+    RETURNS NULL ON NULL INPUT;
+CREATE OR REPLACE FUNCTION TrimmedCenterline(polyline geometry) RETURNS geometry AS '
+    WITH tbla AS (
+        SELECT polyline, (ST_Dump(polyline)).geom as line
+    ),
+    tblb AS (
+        SELECT polyline, line as shortestBranchLine
+        FROM tbla
+        WHERE CountDisconnectedEndpoints(polyline, line) > 0
+        ORDER BY ST_Length(line) ASC
+        LIMIT 1
+    ),
+    tblc AS (
+        SELECT ST_LineMerge(ST_Difference(polyline, shortestBranchLine)) as polyline
+        FROM tblb
+    )
+    SELECT TrimmedCenterline(polyline) as polyline
+    FROM tblc
+    WHERE ST_NumGeometries(polyline) > 1
+    UNION ALL
+    SELECT polyline FROM tblc
+;'
+    LANGUAGE SQL
+    IMMUTABLE
+    RETURNS NULL ON NULL INPUT;
+CREATE OR REPLACE FUNCTION LabelCenterline(input geometry) RETURNS geometry AS '
+    WITH tbla AS (
+        SELECT input as polygon WHERE ST_GeometryType(input) = ''ST_Polygon''
+        UNION ALL
+        SELECT ST_ConcaveHull(input, 0.1) as polygon WHERE ST_GeometryType(input)=''ST_MultiPolygon''
+    ),
+    tblb AS (
+        SELECT ST_MakePolygon(ST_ExteriorRing(polygon)) as polygon
+        FROM tbla
+    ),
+    tblc AS (
+        SELECT polygon, (ST_Dump(ST_VoronoiLines(ST_LineInterpolatePoints(ST_Boundary(polygon), 0.0075)))).geom as lines
+        FROM tblb
+        WHERE ST_GeometryType(polygon) = ''ST_Polygon''
+    ),
+    tbld AS (
+        SELECT ST_LineMerge(ST_Collect(lines)) as polyline
+        FROM tblc
+        WHERE ST_Contains(polygon, lines)
+    )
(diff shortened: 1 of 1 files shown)


nvkelso:
Rad!

nyurik:
this is really impressive!  The sql style is a bit different from other files (e.g. using `'`, etc), but no biggie. I will be studying it for a bit for sure

pnorman:
I'll try to give this a detailed review, as it's something I've been interested in.

nyurik:
Among the above examples, I think only this one seems a bit incorrect:
image

sql/LabelCenterline.sql
```diff
@@ -0,0 +1,71 @@
...
+__Parameters:__
+
+- `geometry` inGeometry - A polygon or multipolygon.
+
+__Returns:__ `geometry(multiline)`
+******************************************************************************/
+CREATE OR REPLACE FUNCTION CountDisconnectedEndpoints(polyline geometry, testline geometry)
+    RETURNS integer AS $$
+    BEGIN
+        RETURN ST_NPoints(ST_RemoveRepeatedPoints(ST_Points(polyline)))
```
pnorman, line 15: An explanation of the equation would be good here.

sql/LabelCenterline.sql
```diff
@@ -0,0 +1,71 @@
...
+                    - ST_NPoints(ST_RemoveRepeatedPoints(ST_Points(ST_Difference(polyline, testline))))
+                    - ST_NPoints(testline) + 2;
+    END
+    $$ LANGUAGE plpgsql;
+CREATE OR REPLACE FUNCTION TrimmedCenterline(inPolyline geometry)
+    RETURNS geometry AS $$
+    DECLARE outPolyline geometry;
+    BEGIN
+        WITH tbla AS (
```
pnorman, line 24: More meaningful CTE names would be good.

sql/LabelCenterline.sql
```diff
@@ -0,0 +1,71 @@
...
+        UNION ALL
+        SELECT polyline INTO outPolyline FROM tblc;
+        RETURN outPolyline;
+    END
+    $$ LANGUAGE plpgsql;
+CREATE OR REPLACE FUNCTION LabelCenterline(inGeometry geometry)
+    RETURNS geometry AS $$
+    DECLARE outPolyline geometry;
+    BEGIN
+        WITH tbla AS (
```
pnorman, line 50: Same with more meaningful CTE names

sql/LabelCenterline.sql
```diff
@@ -0,0 +1,71 @@
...
+            SELECT inGeometry as inPolygon WHERE ST_GeometryType(inGeometry) = 'ST_Polygon'
+            UNION ALL
+            SELECT ST_ConcaveHull(ST_Simplify(inGeometry, 25), 0.2) as inPolygon WHERE ST_GeometryType(inGeometry)='ST_MultiPolygon'
```
pnorman, line 53: I'm skeptical about this approach for multipolygons, as it can lead to geoms outside the original area. This might not be an issue for water, but it is for other types of polygons. If the function is suitable for water only, then it should be renamed.

pnorman (changes requested):
As all three functions are simply a single SQL statement and returning the result of that statement, they should be SQL, not plpgsql. This has performance advantages by avoiding the time to start up a plpgsql environment every function call, and allowing inlining.

More comments would be good, as it's not clear what's going on in all steps.

Minor code nits:
- these functions should probably all be STRICT
- caps on `as` vs `AS` is inconsistent

sql/LabelCenterline.sql
quincylvania: Good point!

sql/LabelCenterline.sql
quincylvania: I see your point. I was running into some errors with ST_ConcaveHull anyway, so I just made it calculate a separate centerline for each component polygon. Future work would be to add a parameter giving more control over what to do with multipolygons. There are cases where you'd want one combined centerline, like for an archipelago.

Screenshot 2023-03-05 at 2 13 38 PM

quincylvania:
Okay, I added a number of enhancements to get nicer results in a reasonably fast timeframe. It now takes just 15 seconds to calculate the basic centerlines for the 395 largest lakes of Michigan, down from 40 seconds. Smoothing is now done proportionally to the size of the feature.

Screenshot 2023-03-10 at 1 42 49 PM

Multiple lines over a given length threshold can be included in the output. This allows more accurate labeling at higher zooms.

Screenshot 2023-03-10 at 1 44 08 PM

Holes over a given size can be included in the calculation as well. This takes a little longer to calculate, 22 seconds for these lakes.

Screenshot 2023-03-10 at 1 44 55 PM

More examples:
Screenshot 2023-03-10 at 1 46 03 PM
Screenshot 2023-03-10 at 1 51 02 PM
Screenshot 2023-03-10 at 1 46 53 PM
Screenshot 2023-03-10 at 1 46 30 PM

There as six parameters to customize these behaviors.

PR #442, opened 2024-02-18
Replace spritezero with spreet

HandyMenny:
`spreet `is a command-line tool that creates [sprites](https://en.wikipedia.org/wiki/Spritesheet).

Why spreet over spritezero?
- It supports svg with a PNG image embeedded (See https://github.com/flother/spreet/issues/60#issuecomment-1739274821)
- It's faster and creates smaller sprites
- It's maintened
- It's a self contained binary 

The first is the most important reason, it fixes https://github.com/openmaptiles/openmaptiles/issues/1549, superseding https://github.com/openmaptiles/openmaptiles/pull/1580 and https://github.com/openmaptiles/spritezero/pull/1

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -58,6 +58,18 @@ RUN set -eux ;\
     mv /usr/src/osmborder/build/src/osmborder /build-bin ;\
     mv /usr/src/osmborder/build/src/osmborder_filter /build-bin
 
+# Build SPREET
+FROM rust:1.76 as rust-builder
+ARG SPREET_REPO="https://github.com/flother/spreet"
+ARG SPREET_VERSION="v0.11.0"
+
+RUN set -eux ;\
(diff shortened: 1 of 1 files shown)


lazaa32:
Hi @HandyMenny 
the `spreet` tool seems very good to me. Glad we do not have to use two other forks to get spritezero working.

Could you please create a related PR into OMT, that in [Makefile](https://github.com/openmaptiles/openmaptiles/blob/master/Makefile#L296-L305)  replaces `spritezero` with `spreet` and use the correct order of <INPUT> and <OUTPUT>? Thanks!

HandyMenny:
> Could you please create a related PR into OMT, that in [Makefile](https://github.com/openmaptiles/openmaptiles/blob/master/Makefile#L296-L305) replaces `spritezero` with `spreet` and use the correct order of and ?

Done

lazaa32 (approved):
Thank you very much!

PR #458, opened 2025-06-18
Add clone3-workaround wrapper to support Docker < 20.10.10

romainreignier:
See https://github.com/openmaptiles/openmaptiles/issues/1719

openmaptiles claims to support Docker >= 1.12.3 but GLIBC >= 2.34 we need Docker >= 20.10.10 and openmaptiles/openmaptiles-tools currently uses Debian 12 as base OS that uses GLBC 2.36.

The wrapper is added as an entrypoint and always used.

We could also use it optionally from the openmaptiles/openmaptiles Makefile based on the Docker version detected?


CommanderStorm:
> Docker < 20.10.10

That was released 2021-10-25. Would a better solution for this not be to just bump the documented requirement?

PR #475, opened 2026-08-11
Migrate tools to Python 3.14

MendryX4:
This PR migrates the `openmaptiles-tools` codebase and Docker environment from Python 3.9 to Python 3.14. The tool versions have been bumped accordingly.

To ensure the migration didn't break the pipeline, the following local tests were successfully executed:
- Passed all unit and integration tests via `make test` using the rebuilt Docker image.
- Ran an end-to-end tile generation using `quickstart.sh`. The data was successfully downloaded, imported into PostGIS, processed by the SQL scripts, and a valid `tiles.mbtiles` file was generated without any errors or warnings.


openmaptiles/__init__.py
```diff
@@ -1 +1 @@
-__version__ = '0.0.0'
+__version__ = '8.0.0'
```
TomPohys:
```suggestion
__version__ = '0.0.0'
```

Please keep just a `0.0.0` so we do not have dependency before release. There is automatic replacement based on tag: https://github.com/openmaptiles/openmaptiles-tools/blob/a7080f2b1a8bf47020716d37408ce04f8c7cb473/.github/workflows/push.yml#L46

TomPohys (commented):
Awesome! Thanks a lot. It looks good!

